PR #3, opened 2022-02-03
Solution

Krimat:


Oleksandr563487953 (approved):

tests/test_main.py
```diff
@@ -1 +1,22 @@
-# write your code here
+from random import randint
+
+
+from app.main import get_human_age
+
+
+def test_first_human_year():
+    cat_years = randint(0, 14)
```
IvanRamyk, line 8: It's better not to use random in tests. Just choose concrete values from this range

tests/test_main.py
```diff
@@ -1 +1,22 @@
...
+    dog_years = randint(0, 14)
+
+    assert get_human_age(cat_years, dog_years) == [0, 0]
+
+
+def test_second_human_year():
```
IvanRamyk, line 14: Same

IvanRamyk (changes requested):

MasterpieceElbow (approved):

PR #5, opened 2022-02-03
Solution

Abnormaltype:


Oleksandr563487953 (approved):

tests/test_main.py
```diff
@@ -1 +1,40 @@
...
+def test_when_cat_age_or_dog_age_is_negative(cat_age, dog_age, expected):
+    assert get_human_age(cat_age, dog_age) == expected
+
+
+@pytest.mark.parametrize(
+    'cat_age, dog_age, expected',
+    [
+        (100, 100, [21, 17]),
+        (200, 200, [46, 37]),
+        (333, 222, [79, 41])
```
IvanRamyk, line 23: It's better not to add tests that check the same scenario.
But it is still ok for now.
Just for future tasks.

IvanRamyk (approved):
Well done!

tests/test_main.py
Abnormaltype: thx for your advice

PR #8, opened 2022-02-03
'Solution'

cloudprogramm:


Oleksandr563487953 (approved):

tests/test_main.py
```diff
@@ -1 +1,10 @@
-# write your code here
+from app.main import get_human_age
+
```
IvanRamyk, line 2: Please, add tests for each class of equivalents:
* when age is under 15
* when age is between 15 and 24
* when age is greater than 24

IvanRamyk (changes requested):

tests/test_main.py
```diff
@@ -1 +1,12 @@
...
+
+def test_right_convert_in_human_age():
+    assert get_human_age(100, 100) == [21, 17]  # checking the differences between large numbers
```
IvanRamyk, line 5: You should create a **separate test** for each class of equivalents. For example:
```
def test_should_return_zero_when_age_is_equal_to_zero():
     assert get_human_age(0, 0) == [0, 0]
     assert get_human_age(14, 14) == [0, 0]


def test_should_return_one_when_age_is_between_15_and_24():
     assert get_human_age(15, 15) == [1, 1]
     assert get_human_age(24, 24) == [1, 1]
```

IvanRamyk (changes requested):

tests/test_main.py
IvanRamyk: could you add tests for `24` and `25` values

IvanRamyk (changes requested):

tests/test_main.py
IvanRamyk: It's important to add tests with values between 15 and 28.
Please, check also all boundary values

IvanRamyk (changes requested):

PR #10, opened 2022-02-03
Develop

denyslipin:


tests/test_main.py
```diff
@@ -1 +1,29 @@
-# write your code here
+import random
+
+from app.main import get_human_age
+
+
+def test_should_return_zeros_if_both_ages_less_than_15():
+    cat = random.randint(0, 14)
```
IvanRamyk, line 7: It's better not to use random where it is unnecessary.
Just pick some values from this range by yourself

IvanRamyk (changes requested):

IvanRamyk (approved):
Well done!

PR #12, opened 2022-02-03
Solution

vldtsw:


Oleksandr563487953 (approved):

tests/test_main.py
```diff
@@ -1 +1,17 @@
...
+def test_return_zeroes_if_animals_age_equal_to_0():
+    assert get_human_age(0, 0) == [0, 0]
+
+
+def test_return_zeros_if_animals_age_less_then_15():
+    assert get_human_age(5, 1) == [0, 0]
+    assert get_human_age(14, 14) == [0, 0]
+
+
+def test_animals_age_more_then_15():
```
IvanRamyk, line 13: Could you split this test into several for each class of equivalence?
Separate tests when the answer is one, two, and greater.

IvanRamyk (changes requested):

tests/test_main.py
```diff
@@ -1 +1,16 @@
...
+    assert get_human_age(15, 15) == [1, 1]
+    assert get_human_age(23, 23) == [1, 1]
+
+
+def test_two_years():
+    assert get_human_age(24, 24) == [2, 2]
+
+
+def test_second_year_and_greater():
+    assert get_human_age(30, 30) == [3, 3]
```
MasterpieceElbow, line 16: Add test when function returns different values when it takes same values

MasterpieceElbow (approved):

tests/test_main.py
MasterpieceElbow, line 16: Add test when function returns different values when it takes same values

MasterpieceElbow (changes requested):

IvanRamyk (approved):

PR #14, opened 2022-02-03
Solution

alix-mak:


liuda-chuzdiuk (approved):

tests/test_main.py
```diff
@@ -1 +1,88 @@
-# write your code here
+from app.main import get_human_age
+
+
+def test_output_type():
+    assert type(get_human_age(1, 1)) == list
+
+    is_int = True
+    if type(get_human_age(1, 1)) == list:
```
MasterpieceElbow, line 8: You don't need that condition, you assert that it is a list above.

tests/test_main.py
```diff
@@ -1 +1,88 @@
...
+        for i in get_human_age(30, 31):
+            if not type(i) == int:
+                is_int = False
+                break
+
+    assert is_int
```
MasterpieceElbow, line 14: It can be more shorted with `assert all([isinstanse(value, int) for value in get_human_age(30, 31)])`

tests/test_main.py
```diff
@@ -1 +1,88 @@
...
+def test_age_less_than_15():
+    expected = {
+        (2, 2): [0, 0],
+        (10, 10): [0, 0],
+        (14, 14): [0, 0]
+    }
+    values = [(2, 2), (10, 10), (14, 14)]
+
+    actual = {item[1]: get_human_age(values[item[0]][0], values[item[0]][1])
+              for item in enumerate(values)}
```
MasterpieceElbow, line 30: You can include zero age here and rename test to `test_first_age` because you testing that ages less than 1.
Use `mark.parametrize` here and in other tests.

tests/test_main.py
```diff
@@ -1 +1,88 @@
...
+        (0, 24): [0, 2],
+        (0, 29): [0, 3],
+        (0, 114): [0, 20]
+    }
+    values = [(0, 24), (0, 29), (0, 114)]
+
+    actual = {item[1]: get_human_age(values[item[0]][0], values[item[0]][1])
+              for item in enumerate(values)}
+
+    assert actual == expected
```
MasterpieceElbow, line 74: You don't need to test dog and cat ages separately. Use `mark.parametrize`.

MasterpieceElbow (changes requested):

tests/test_main.py
alix-mak: Used pytest.mark.parametrize

MasterpieceElbow (approved):

PR #15, opened 2022-02-03
Solution

nikon4uk:


8Men5Camels (approved):

tests/test_main.py
```diff
@@ -1 +1,21 @@
...
+
+def test_first_fifteen_years_for_cat():
+    assert get_human_age(15, 14) == [1, 0]
+
+
+def test_first_fifteen_years_for_dog():
+    assert get_human_age(14, 15) == [0, 1]
+
+
+def test_next_nine_years_for_pets():
```
IvanRamyk, line 12: Could you also create a test for boundary value [23, 23]?

tests/test_main.py
```diff
@@ -1 +1,21 @@
...
+    assert get_human_age(24, 24) == [2, 2]
+
+
+def test_every_four_next_years_for_cat():
+    assert get_human_age(28, 28) == [3, 2]
+
+
+def test_every_five_next_years_for_dog():
```
IvanRamyk, line 20: Could you add the test case for 4+ years?

IvanRamyk (changes requested):

IvanRamyk (approved):

PR #17, opened 2022-02-03
Solution

phantom-terminal:


8Men5Camels (approved):

tests/test_main.py
```diff
@@ -1 +1,26 @@
-# write your code here
+from app.main import get_human_age
+
+
+def test_age_should_be_positive_and_round():
```
IvanRamyk, line 4: Could you create two separate tests for these cases?

IvanRamyk (changes requested):

tests/test_main.py
phantom-terminal: Originally this test was a section of two tests. But I thought they could be combined. I'll split it back up.

MasterpieceElbow (approved):

PR #20, opened 2022-02-03
Solution

TPolina:


Oleksandr563487953 (approved):

tests/test_main.py
```diff
@@ -1 +1,23 @@
-# write your code here
+from app.main import get_human_age
+
+
+def test_cat_age_converted_to_human_age_correctly():
```
IvanRamyk, line 4: Please create a test for each case (class of equivalence)

IvanRamyk (changes requested):

tests/test_main.py
TPolina: Did you mean that I should write separate test function for each assertion? Is it really necessary?

MasterpieceElbow (approved):

PR #22, opened 2022-02-07
'Solution'

Zikkolion:


Oleksandr563487953 (approved):

tests/test_main.py
```diff
@@ -1 +1,16 @@
-# write your code here
+import pytest
+from app.main import get_human_age
+
+
+@pytest.mark.parametrize(
+    'cat_age, dog_age, result',
+    [
+        [0, 0, [0, 0]],
```
IvanRamyk, line 8: Could you add the `id` value via `pytest.params` to provide a description for each test?

IvanRamyk (changes requested):

PR #35, opened 2022-06-01
Solution

KatiukhaO:


app/test_main.py
```diff
@@ -1,3 +1,38 @@
+import pytest
+
 from app.main import get_human_age
 
-# write your code here
+
+@pytest.mark.parametrize("arg1, arg2", [(15, 15), ])
```
rotsen18, line 6: If I only had a list of single arguments I would not use parametrization.

app/test_main.py
```diff
@@ -1,3 +1,38 @@
...
+def test_len_list_return_get_human_age(arg1, arg2):
+    assert len(get_human_age(arg1, arg2)) == 2
+
+
+@pytest.mark.parametrize("args", [(15, 15)])
+def test_type_return_value_in_list_get_human_age(args):
+    f = True
+    for i in range(len(get_human_age(*args))):
+        f = isinstance(get_human_age(*args)[i], int)
+    assert f
```
rotsen18, line 21:
```suggestion
    assert = all(isinstance(i, int) for i in get_human_age(*args))
```
if you need check output types maybe simplify a little ;)

rotsen18 (approved):

app/test_main.py
MasterpieceElbow, line 6: As Taras said. don't use `mark.parametrize` for only one case.

app/test_main.py
MasterpieceElbow, line 21: Taras is right, you can make it more concise. Also, `f` and `i` are not good variable names.

MasterpieceElbow (changes requested):

app/test_main.py
```diff
@@ -1,3 +1,29 @@
...
 
-# write your code here
+
+def test_type_return_get_human_age():
+    assert isinstance(get_human_age(15, 15), list)
+
+
+@pytest.mark.parametrize("args", [(15, 15), (0, 0)])
+def test_type_return_value_in_list_get_human_age(args):
+    assert all(isinstance(i, int) for i in get_human_age(*args))
```
MasterpieceElbow, line 12: Don't use `i` as a variable name.

MasterpieceElbow (approved):

PR #36, opened 2022-06-02
'Solution'

Aleksei-Isaev:


Aleksei-Isaev:
> What about very old anymals, let's say get_human_age(100, 100)?
If we talking about parrots or turtles, it may happen)

MasterpieceElbow (approved):

PR #54, opened 2022-06-04
Solution

Igor-Cegelnyk:


IvanRamyk (changes requested):
You should write tests for boundary values and different equivalence classes

IvanRamyk (approved):
Good job!

PR #57, opened 2022-06-05
Solution

Evgeniy994:


app/test_main.py
```diff
@@ -1,3 +1,29 @@
 from app.main import get_human_age
 
-# write your code here
+
```
Aleksei-Isaev, line 3: In my opinion, there are too many tests, but if it works, it's up to you.

Aleksei-Isaev (approved):

MasterpieceElbow (approved):

PR #62, opened 2022-06-06
'Solution'

YuraBeseda:


app/test_main.py
```diff
@@ -1,3 +1,21 @@
...
+
+def test_if_age_is_more_than_15():
+    assert get_human_age(23, 23) == [1, 1]
+
+
+def test_if_age_is_more_than_24():
+    assert get_human_age(27, 27) == [2, 2]
+
+
+def test_if_age_is_very_old():
```
Chardona, line 20: "very old" is  an ambiguous name. Recommend changing to something more specific.

Chardona (approved):

TPolina (approved):

PR #69, opened 2022-06-21
Solution

Splend1ed:


app/test_main.py
```diff
@@ -1,3 +1,5 @@
 from app.main import get_human_age
 
-# write your code here
+
+def test_cat_and_god_years_convert():
```
IvanRamyk, line 4: You should provide more tests for boundary values and classes of equivalence

IvanRamyk (changes requested):

app/test_main.py
```diff
@@ -1,3 +1,27 @@
...
+                15, 15,
+                [1, 1],
+                id="should return 1 human year for first animal years"
+            ),
+            pytest.param(
+                24, 24,
+                [2, 2],
+                id="should add `extra year` every 9 animal next years"
+            ),
+            pytest.param(
```
IvanRamyk, line 19: Could you add a test [28, 28] to make sure that the numbers of extra years are really different

app/test_main.py
```diff
@@ -1,3 +1,27 @@
...
+class TestCatDogYears:
+    @pytest.mark.parametrize(
+        "cat_years, dog_years, human_years",
+        [
+            pytest.param(
+                15, 15,
+                [1, 1],
+                id="should return 1 human year for first animal years"
+            ),
+            pytest.param(
```
IvanRamyk, line 14: Could you add a test [23, 23] to make sure that the function uses exactly a 9-year difference?

IvanRamyk (changes requested):

IvanRamyk (approved):
Good job!

PR #73, opened 2022-07-20
Solution

DHushchyk:


app/test_main.py
```diff
@@ -1,3 +1,13 @@
 from app.main import get_human_age
 
-# write your code here
+
```
IvanRamyk, line 3: Please, write tests for the boundary values

IvanRamyk (changes requested):

IvanRamyk (approved):
Well done!

PR #74, opened 2022-07-20
Solution

DanRoman-code:


app/test_main.py
```diff
@@ -1,3 +1,7 @@
 from app.main import get_human_age
 
-# write your code here
+
+def test_of_correctly_type_and_result():
```
IvanRamyk, line 4: Please, add tests for the boundary values

IvanRamyk (changes requested):

IvanRamyk (approved):
Good job!

PR #77, opened 2022-07-22
solution

tarkaiev:


Abnormaltype (changes requested):
also, add tests about negative input, zeros, and large numbers

Abnormaltype (approved):
Good job!

PR #79, opened 2022-07-26
Solution

andrii-padalko:


Abnormaltype (changes requested):
add tests for these cases: negative input, large input

Abnormaltype (approved):
Well done!

PR #81, opened 2022-08-02
Solution

Vlad-Peresta:


app/test_main.py
```diff
@@ -1,3 +1,22 @@
+import pytest
+
 from app.main import get_human_age
 
-# write your code here
+
+@pytest.mark.parametrize("cat_age,dog_age,result_in_year",
+                         [
```
IvanRamyk, line 7: Please, improve your formatting. It's a bit unreadable know

IvanRamyk (changes requested):

app/test_main.py
Vlad-Peresta: ok, but I don't know how to write parameters in first argument of parametrize (there should be no indents) more readable

IvanRamyk: https://pastebin.com/8yYuSNfa
I showed how you can make this code more readable, as for me

IvanRamyk (changes requested):

app/test_main.py
Vlad-Peresta: I agree this style of code is more readable. Thank you very mach!)

IvanRamyk (approved):
Good job!

PR #85, opened 2022-08-12
Solution

ansicat:


Abnormaltype (changes requested):
also, add tests when the function takes negative values

Abnormaltype (approved):
Well done!

PR #86, opened 2022-08-13
cat_and_dog_years_v1

Povshok:


app/test_main.py
```diff
@@ -1,3 +1,17 @@
...
+def test_should_return_zero_when_ages_less_15():
+    assert get_human_age(0, 14) == [0, 0]
+
+
+def test_should_return_1_when_ages_between_15_and_23():
+    assert get_human_age(15, 23) == [1, 1]
+
+
+def test_should_add_1_year_every_4_pets_years_when_ages_24_and_more():
+    assert get_human_age(100, 28) == [21, 2]
```
DmytroSvirsa, line 13: it would be better to check the boundary cases like the 3-rd and 2-nd years of cat_and_dog_years life like you did for a 1-st year, cower from both sides - (14=> 0, 15 => 1)

app/test_main.py
```diff
@@ -1,3 +1,17 @@
...
+
+
+def test_should_return_zero_when_age_equal_0():
```
DmytroSvirsa, line 16: The name of the test does not reflect the meaning of the test you checking if less than 0

DmytroSvirsa (changes requested):

IvanRamyk (approved):
Good job!

PR #88, opened 2022-08-20
Solution

UaMermaid:


Abnormaltype (changes requested):
add test when a function takes negative values

Abnormaltype (approved):
Well done!

PR #98, opened 2022-08-29
Solution

VladyslavKlevanskyi:


app/test_main.py
```diff
@@ -1,3 +1,37 @@
...
+        ]
+
+    )
+    def test_calculating_ages_correctly(
+            self,
+            cat_age,
+            dog_age,
+            human_age
+    ):
+        assert get_human_age(cat_age, dog_age) == human_age
```
IvanRamyk, line 37: please, check all boundary values

IvanRamyk (changes requested):

app/test_main.py
VladyslavKlevanskyi: Did you mean 0 and 100?

IvanRamyk: I mean (14-15), (27-28-29), etc

IvanRamyk (changes requested):

IvanRamyk (approved):
Good job!

PR #99, opened 2022-08-29
'Solution'

Kev1nXD:


app/test_main.py
```diff
@@ -1,3 +1,37 @@
...
+            ),
+            pytest.param(
+                23,
+                23,
+                [1, 1],
+                id="should return 1 if cat's and dog's age >= 15 or <= 23"
+            )
+        ]
+    )
+    def test_get_human_should_work_properly(self,
```
IvanRamyk, line 33: please, write tests for boundary values, such as 14-15, 23-24-25, etc

IvanRamyk (changes requested):

app/test_main.py
IvanRamyk: what about other boundary values?

IvanRamyk (changes requested):

IvanRamyk (approved):
Good job!

PR #103, opened 2022-09-10
Test for cat and dog years

Foggy-Cat:


Abnormaltype (changes requested):
`pytest` - doesn't use with classes

Abnormaltype (changes requested):
add more tests:
negative input, large input, zeros input

Abnormaltype (approved):
Well done!

PR #104, opened 2022-09-17
'Solution'

Pavilion4ik:


app/test_main.py
```diff
@@ -1,3 +1,26 @@
 from app.main import get_human_age
 
-# write your code here
+
+def test_should_return_correct_result_if_ages_bigger_then_14():
+    cat_age, dog_age = 28, 28
+    assert get_human_age(cat_age, dog_age) == [3, 2]
```
IvanRamyk, line 6:
```suggestion
    assert get_human_age(28, 28) == [3, 2]
```
Do not create extra variables

IvanRamyk (changes requested):

app/test_main.py
```diff
@@ -1,3 +1,21 @@
 from app.main import get_human_age
+import pytest
 
-# write your code here
+
+class Test:
+    @pytest.mark.parametrize(
+        "cat_age, dog_age, result",
+        [
+            (28, 28, [3, 2]),
```
IvanRamyk, line 9: Please add more tests for different boundary values, such as (14, 14), (27, 27) ...

IvanRamyk (changes requested):

DmytroSvirsa (approved):

PR #106, opened 2022-09-17
py_cat_dog_years_Solution

vadim-kosnovskiy:


app/test_main.py
```diff
@@ -1,3 +1,18 @@
 from app.main import get_human_age
 
-# write your code here
+
+def test_equal_right_result():
+    args = (
```
IvanRamyk, line 5: Please, create separate tests for every case and name them to describe each test case

IvanRamyk: It's better practice to create a separate test per case because it's much easier to find an input that leads to an error if some test failed

IvanRamyk (changes requested):

app/test_main.py
```diff
@@ -1,3 +1,29 @@
...
+def test_zero_ages():
+    ages = {"cat_age": 14, "dog_age": 14, "human_age": [0, 0]}
+    assert get_human_age(
+        ages["cat_age"], ages["dog_age"]) == ages["human_age"]
+
+
+def test_first_ages():
+    ages = {"cat_age": 15, "dog_age": 15, "human_age": [1, 1]}
+    assert get_human_age(
+        ages["cat_age"], ages["dog_age"]) == ages["human_age"]
```
IvanRamyk, line 13:
```suggestion
    assert get_human_age(15, 15) == [1, 1]
```
It's better to use explicit values, it makes the code more readable

IvanRamyk (changes requested):

app/test_main.py
```diff
@@ -1,3 +1,21 @@
...
+def test_first_ages():
+    assert get_human_age(15, 15) == [1, 1]
+
+
+def test_second_ages():
+    assert get_human_age(24, 24) == [2, 2]
+
+
+def test_more_ages():
+    assert get_human_age(28, 29) == [3, 3]
```
IvanRamyk, line 17: Write tests for more boundary values: 
(28, 28)
(23, 23)

IvanRamyk (changes requested):

IvanRamyk (approved):
Good job!

PR #109, opened 2022-09-22
Solution

wQuelS:


sbublinskyi (approved):
not bad )
you also may use `@pytest.mark.parametrize` in order not to duplicate the code

wQuelS:
Yeah, i've started using @pytest.mark.parametrize in las tasks)

Nattalli (approved):
it's really would be better to use a pytest, but anyway it looks good!

PR #114, opened 2022-09-24
Solution

IvanStored:


wQuelS (changes requested):
I believe u have to use id in pytest.param to provide more info in case some tests are not passed

wQuelS (approved):
GJ

MasterpieceElbow (approved):
GJ!

PR #115, opened 2022-09-25
Solution

mltdll:


app/test_main.py
```diff
@@ -1,3 +1,58 @@
...
+            28,
+            [2, 2],
+            id="should return [2, 2] when inputs are "
+               "one less than 3 human years"
+        ),
+        pytest.param(
+            28,
+            29,
+            [3, 3],
+            id="should return [3, 3] when inputs are "
```
wQuelS, line 52: check your id description, cause your're testing function that takes cat's/dog's years as input and returns human age. Now it is a bit hard to understand what is wrong if some tests are failed.

app/test_main.py
```diff
@@ -1,3 +1,58 @@
...
+            24,
+            [2, 2],
+            id="should return [2, 2] when inputs are exactly 2 human years"
+        ),
+        pytest.param(
+            27,
+            28,
+            [2, 2],
+            id="should return [2, 2] when inputs are "
+               "one less than 3 human years"
```
wQuelS, line 46: what is this phrase even supposed to mean?

app/test_main.py
```diff
@@ -1,3 +1,58 @@
...
+        pytest.param(
+            14,
+            14,
+            [0, 0],
+            id="should return [0, 0] when inputs are "
+               "one less than 1 human year"
+        ),
+        pytest.param(
+            15,
+            15,
```
wQuelS, line 24: also, consider to group test cases for 15-23 human years.

wQuelS (changes requested):

app/test_main.py
mltdll: This phrase is supposed to mean that the test covers the case where both inputs are exactly one year short of reaching 3 human years, i.e. the upper boundary of the third equivalence class.
I'll see what I can do to make it easier to understand.

app/test_main.py
mltdll: Group them? I'm not sure I even have a single one.
Though it would make sense to add one to check if the behavior persists for larger inputs.

wQuelS (approved):
Really good that u've decided to add check for large numbers. By grouping i meant to make only one test for 15-23 cat/dog years. But anyway, your variant looks nice too:)

Nattalli (approved):
Good job! 😸

PR #118, opened 2022-09-26
Solution

StasOkhrym:


mltdll (approved):
Seems good to me.

app/test_main.py
```diff
@@ -1,3 +1,55 @@
...
+                id="should return correct age with random numbers"
+            ),
+            pytest.param(
+                45, 18, [7, 1],
+                id="should return correct age with random numbers"
+            )
+        ]
+    )
+    def test_get_correct_age(
+            self,
```
Nattalli, line 50: I think that `self` is unneeded here

app/test_main.py
```diff
@@ -1,3 +1,55 @@
...
+                -1, -1, [0, 0],
+                id="should return 0 if animal age is negative"
+            ),
+            pytest.param(
+                100, 100, [21, 17],
+                id="should work correct with big numbers"
+            ),
+            pytest.param(
+                24, 39, [2, 5],
+                id="should return correct age with random numbers"
```
Nattalli, line 37: can you change the id? It should be unique for each case

Nattalli (changes requested):

Nattalli (approved):
Well done!

PR #119, opened 2022-09-27
Solution

AVyha:
I think my tests enough for this function, i can't imagine what test i can add.

### DIFF
--- a/app/test_main.py
+++ b/app/test_main.py
@@ -1,3 +1,41 @@
 from app.main import get_human_age
+import pytest
 
-# write your code here
+
+@pytest.mark.parametrize(
+    "cat_age,dog_age,expected",
+    [
+        (
+            28,
+            28,
+            [3, 2]
+        ),
+        (
+            0,
+            0,
+            [0, 0]
+        ),
+        (
+            15,
+            15,
+            [1, 1]
+        ),
+        (
+            100,
+            100,
+            [21, 17]
+        )
+    ]
+)
+def test_function_correct_working(cat_age, dog_age, expected):
+    assert get_human_age(cat_age, dog_age) == expected
+
+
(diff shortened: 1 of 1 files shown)


mltdll (changes requested):
In ideal scenario, you should test for all the boundaries between different outputs, for example, for cat years:
- 0;
- 14: largest possible value that returns 0 human years;
- 15: smallest possible value that returns 1 human years;
- 23: largest possible value that returns 1 human years; and so on.

Also, you could theoretically add negative testing, e.g. for negative numbers or other types, but instructions don't mention function behavior for cases like that, so it doesn't make sense in this task.

AVyha:
> In ideal scenario, you should test for all the boundaries between different outputs, for example, for cat years:
> 
> * 0;
> * 14: largest possible value that returns 0 human years;
> * 15: smallest possible value that returns 1 human year;
> * 23: largest possible value that returns 1 human year; and so on.
> 
> UPD: This does not apply to numbers past 28, since they increase linearly, and you can just test for a large value.
> 
> Also, you could theoretically add negative testing, e.g. for negative numbers or other types, but instructions don't mention function behavior for cases like that, so it doesn't make sense in this task.
> 
> Also, consider adding `id` to your test cases so that they have meaningful titles.

Ok, thank you for this answer, i forget to add id)

mltdll (approved):
I still think it would make sense to add tests for 14, 23, etc, but fine.

AVyha:
I think it does not matter that test takes lower value like `0, 0` or `15, 15` instead `14, 14` or `23, 23` but its not hard for me to add it) Thank you for comments.

mltdll (approved):

Nattalli (approved):
Good!

PR #121, opened 2022-09-28
Solution

SorokaIgor:


app/test_main.py
```diff
@@ -1,3 +1,16 @@
 from app.main import get_human_age
 
-# write your code here
+
+def test_zero_ages():
+    result1 = get_human_age(15, 15)
+    assert result1 == [1, 1]
```
DmytroSvirsa, line 6: test edge situation from both sides

app/test_main.py
```diff
@@ -1,3 +1,16 @@
...
+
+
+def test_first_ages():
+    result2 = get_human_age(24, 24)
```
DmytroSvirsa, line 10: test edge situation from both sides

app/test_main.py
```diff
@@ -1,3 +1,16 @@
...
+    assert result2 == [2, 2]
+
+
+def test_second_ages():
+    result = get_human_age(28, 29)
```
DmytroSvirsa, line 15: test edge situation from both sides

DmytroSvirsa (commented):
also add test for correct count of big numbers

DmytroSvirsa (changes requested):
read comments

app/test_main.py
```diff
@@ -1,3 +1,17 @@
 from app.main import get_human_age
 
-# write your code here
+
+def test_zero_ages():
+    assert get_human_age(16, 14) == [1, 0]
```
DmytroSvirsa, line 5: 16 is not adge variant 15 - yes

app/test_main.py
```diff
@@ -1,3 +1,17 @@
...
+
+
+def test_first_ages():
+    assert get_human_age(24, 24) == [2, 2]
```
DmytroSvirsa, line 9: what about 23 it should be 1 or 2?

app/test_main.py
```diff
@@ -1,3 +1,17 @@
...
+
+
+def test_second_ages():
+    assert get_human_age(28, 28) == [3, 2]
```
DmytroSvirsa, line 13: what about (28, 29), will it be the same?

DmytroSvirsa (changes requested):
to achieve the full edged situation coverage you need to add few more tests

app/test_main.py
DmytroSvirsa: almost everything is done great but this test still have a problem

DmytroSvirsa (changes requested):

DmytroSvirsa (approved):

PR #124, opened 2022-09-28
Solution for the task

olia-trofymchuk:


app/test_main.py
```diff
@@ -1,3 +1,61 @@
...
+                24,
+                24,
+                [2, 2],
+                id="given age is equal to 24"
+            ),
+            pytest.param(
+                27,
+                27,
+                [2, 2],
+                id="given age is under cat and dog limitation"
```
mltdll, line 43: Id of this test doesn't seem very clear to me. Consider changing it.

app/test_main.py
```diff
@@ -1,3 +1,61 @@
...
+            ),
+            pytest.param(
+                28,
+                28,
+                [3, 2],
+                id="given age return plus year to cat"
```
mltdll, line 49: Same here. I don't think picking a name that can't be understood without reading the name of previous test is a very good idea.

mltdll (changes requested):

app/test_main.py
olia-trofymchuk: I agree with you. It's better to change it. Thank you!

app/test_main.py
```diff
@@ -1,3 +1,61 @@
 from app.main import get_human_age
+import pytest
 
-# write your code here
+
+class TestPetAgeInHumanYears:
```
mltdll, line 5: By the way, you don't need to introduce a class if it's not for test categorization. And since you only have one test... You clearly don't need it if you don't intend to extend the test suite later.
I'm not sure if that's an issue, though.

mltdll (approved):
Seems ok to me.

PR #126, opened 2022-09-29
Solution

hai-le-di:


app/test_main.py
```diff
@@ -1,3 +1,21 @@
...
+def test_should_return_zeros_if_age_less_fifteen():
+    assert get_human_age(14, 14) == [0, 0]
+
+
+def test_should_return_one_if_age_equals_fifteen():
+    assert get_human_age(15, 15) == [1, 1]
+
+
+def test_twenty_four_years_should_convert_into_two_human_years():
+    assert get_human_age(24, 24) == [2, 2]
```
DmytroSvirsa, line 17: test the top border the same as you done for [0, 0]

app/test_main.py
```diff
@@ -1,3 +1,21 @@
...
+
+
+def test_twenty_eight_and_twenty_nine_should_convert_into_three():
+    assert get_human_age(28, 29) == [3, 3]
```
DmytroSvirsa, line 21: test if calculate correct for bigger values

DmytroSvirsa (changes requested):

app/test_main.py
hai-le-di: What's the top border?

PR #131, opened 2022-10-01
Solution

Andriy-Makarenko:


app/test_main.py
```diff
@@ -1,3 +1,36 @@
+import pytest
+
 from app.main import get_human_age
+from typing import List
```
mltdll, line 4: Imports from standard library should come first, then third-party packages, and then local imports. All categories should be separated by an extra newline.

mltdll (changes requested):

app/test_main.py
```diff
@@ -1,3 +1,37 @@
+import pytest
+
+from typing import List
```
mltdll, line 3: I think typing should be higher than pytest, since the latter is a third-party library. By the way, if you are running Python 3.10, you can annotate `list[int]` instead of `typing.List[int]`.

mltdll (changes requested):

mltdll (approved):

app/test_main.py
```diff
@@ -1,3 +1,37 @@
...
+            id="Should return 1 if cat's or dog's age 15-23"
+        ),
+        pytest.param(
+            24, 24, [2, 2],
+            id="Should return 2 if cat's age 24-27 and dog's age 24-28"
+        ),
+        pytest.param(
+            28, 29, [3, 3],
+            id="Should return 3 if cat's age 28-31 and dog's age 29-33"
+        ),
```
TPolina, line 30: you don't test edge cases here which is the most important. What if the developer forgot that cat and dog years count differently? If dog_age 28 will be the result 2 or maybe 3 as well? Also, you should add some big numbers

TPolina (changes requested):

TPolina (approved):
Well done!

PR #132, opened 2022-10-01
'Solution'

Yurchikkk:


app/test_main.py
```diff
@@ -1,3 +1,37 @@
...
+    result = get_human_age(24, 24)
+    assert result == [2, 2]
+
+
+def test_should_return_correct_age_if_animal_age_is_more_than_24():
+    result = get_human_age(100, 100)
+    assert result == [21, 17]
+
+
+def test_should_raise_error_if_any_of_animal_age_is_less_than_0():
```
Abnormaltype, line 35: `age_is_less_than_0` it is correct?)

Abnormaltype (changes requested):

app/test_main.py
Yurchikkk: sorryyy, fixed.
I also wanted to raise an error when we get negative animal_age , but function in ok with that, should I change function to write this test or it is not important now?

Abnormaltype: you just can add this test) It's not so important

Abnormaltype (approved):
Well done!

PR #135, opened 2022-10-02
Solution

rostyak:


app/test_main.py
```diff
@@ -1,3 +1,25 @@
...
+
+def test_should_return_correct_output_for_different_ages():
+    assert [1, 2] == get_human_age(23, 24)
+
+
+def test_should_return_different_result_for_28_pets_ages():
+    assert [3, 2] == get_human_age(28, 28)
+
+
+def test_for_extra_ages():
```
mltdll, line 20: I think "larger" would be a better description for those ages than "extra" is. However, not sure if it's a problem.

mltdll (changes requested):
Requesting changes so that you are guaranteed to see the message. Just re-request review if you disagree with suggestion.

mltdll (approved):

app/test_main.py
```diff
@@ -1,3 +1,25 @@
 from app.main import get_human_age
 
-# write your code here
+
+def test_should_return_zeros_when_ages_equal_0():
+    assert [0, 0] == get_human_age(0, 0)
```
Y-Havryliv, line 5: change order please `get_human_age(0, 0) == assert [0, 0]`

Y-Havryliv: also, it would be great to use `pytest.mark.parametrize`

Y-Havryliv (changes requested):

Y-Havryliv (approved):
GJ!

PR #138, opened 2022-10-03
Complete!

ketstap162:


app/test_main.py
```diff
@@ -1,3 +1,10 @@
 from app.main import get_human_age
 
-# write your code here
+
+def test_should_return_expected_value():
```
mltdll, line 4: You should either use `pytest.mark.parametrize()` or create a function for each test. Either way, it is necessary to provide short explanations for all the tests, either in id or in the function name.

P.S. Huh. That's a lot of eithers.

mltdll (changes requested):

mltdll (approved):

Y-Havryliv (approved):
GJ!

PR #142, opened 2022-10-04
Solution)

Gr1zzZzzly:


app/test_main.py
```diff
@@ -1,3 +1,40 @@
 from app.main import get_human_age
+import pytest
 
-# write your code here
+
+class TestDogAge:
```
TPolina, line 5: why did you decide to put these tests in a class?

app/test_main.py
```diff
@@ -1,3 +1,40 @@
...
+            (14, 0),
+            (15, 1),
+            (23, 1),
+            (24, 2),
+            (27, 2),
+            (28, 3),
+            (100, 21)
+        ]
+    )
+    def test_for_dog(self, age: int, result: int) -> None:
```
TPolina, line 38: you don't need to create separate tests for cats and dogs. You test one function so test it comprehensively

TPolina (changes requested):

app/test_main.py
Gr1zzZzzly: cause i think if we will make more tests for this func , separate classes is right 
It's more comfortable for me)

app/test_main.py
```diff
@@ -1,3 +1,21 @@
...
+            (14, 14, [0, 0]),
+            (15, 15, [1, 1]),
+            (23, 23, [1, 1]),
+            (24, 24, [2, 2]),
+            (27, 27, [2, 2]),
+            (28, 29, [3, 3]),
+            (100, 100, [21, 17])
+        ]
+    )
+    def test_for_dog(self, age_dog: int, age_cat: int, result: list) -> None:
```
MasterpieceElbow, line 19: Not a relevant name if you test both dog and cat.

MasterpieceElbow (changes requested):

TPolina (approved):

PR #144, opened 2022-10-04
Solution

VadymShkarbul:


app/test_main.py
```diff
@@ -1,3 +1,40 @@
...
+            24, 24, [2, 2],
+            id="Return [2,2] if animals ages are greater than 23"
+        ),
+        pytest.param(
+            28, 28, [3, 2],
+            id="Return [3, 2] if animals ages are 28"
+        ),
+        pytest.param(
+            100, 100, [21, 17],
+            id="Should return [100, 100] if animals ages are 100"
```
HStakhiv, line 35: `[100, 100]`? Are you sure?

HStakhiv (changes requested):

app/test_main.py
VadymShkarbul: Yes you are right! I'll fix it :)

Y-Havryliv (approved):
GJ!

PR #147, opened 2022-10-04
'Solution'

Menselface:


app/test_main.py
```diff
@@ -1,3 +1,17 @@
 from app.main import get_human_age
+import pytest
 
-# write your code here
+
+class TestHumanAge:
```
Gr1zzZzzly, line 5: i think u must implement 2 separate tests for dog and cat, cause it's not obvious where is error

Gr1zzZzzly (changes requested):

Y-Havryliv (approved):
gj, but why did you use a class?

PR #149, opened 2022-10-04
Solution

matechris:


app/test_main.py
```diff
@@ -1,3 +1,66 @@
+import pytest
+
 from app.main import get_human_age
 
-# write your code here
+
+class TestHumanAge:
```
HStakhiv, line 6: I think this class is useless because you have only one category and one test function

HStakhiv (changes requested):

HStakhiv (approved):
Good

Y-Havryliv (approved):
GJ!

PR #153, opened 2022-10-05
'Solution'

OvDima1:


app/test_main.py
```diff
@@ -1,3 +1,34 @@
 from app.main import get_human_age
 
-# write your code here
+
+def test_should_return_0_when_age_is_0():
+    assert get_human_age(0, 0) == [0, 0]
+
+
+def test_type_of_age():
```
Vasyl-Poremchuk, line 8: I think you do not need to check the data type because you should have annotations in the module `main.py` that assume the input data is the same as the data written in the annotations. You can simpy test the function with these values `get_human_age(15, 15)`

app/test_main.py
```diff
@@ -1,3 +1,34 @@
 from app.main import get_human_age
 
-# write your code here
+
+def test_should_return_0_when_age_is_0():
```
Vasyl-Poremchuk, line 4: Add `-> None:` annotation in all functions

Vasyl-Poremchuk (changes requested):

Y-Havryliv (approved):
Well done!

PR #154, opened 2022-10-05
Solution

skrypka-dmytro:


app/test_main.py
```diff
@@ -1,3 +1,40 @@
...
+        "cat_years, dog_years, expended_years",
+        [
+
+            pytest.param(
+                0, 0, [0, 0],
+                id="Should return [0, 0] when years equals zero"
+            ),
+            pytest.param(
+                14, 14, [0, 0],
+                id="Should return [0, 0] when years <= 15 years"
```
mltdll, line 17: There's a typo here. It's either `< 15` or `<= 14`.

app/test_main.py
```diff
@@ -1,3 +1,40 @@
...
+            ),
+            pytest.param(
+                15, 15, [1, 1],
+                id="Should return [1, 1] when age cat equals\
```
mltdll, line 21: And another one here: `age cat` instead of `cat age`. 
By the way, since this string is enclosed in parentheses anyway, you can end string literal here and continue on the next line: 
```python
id="..."
   "..."

File: app/test_main.py [hunk 3]
```diff
@@ -1,3 +1,40 @@
...
+                 15 and dog age equals 15"
+            ),
+            pytest.param(
+                24, 24, [2, 2],
+                id="Should return [2, 2] when age cat equals\
```
mltdll, line 26: Other occurrences of `cat age` also feature the same typo.

mltdll (changes requested):

mltdll (approved):
GJ

PR #158, opened 2022-10-05
Solution

skrypka-dmytro:


mltdll (approved):
GJ, but I think it would be easier to add main.py to excluded files in `.flake8`, since this file's type annotations are not our responsibility. :)

app/test_main.py
```diff
@@ -1,3 +1,43 @@
+import pytest
+
 from app.main import get_human_age
 
-# write your code here
+
+class TestCatAndDogYears:
```
Y-Havryliv, line 6: why did you decide to use a class?

Y-Havryliv (changes requested):

app/test_main.py
skrypka-dmytro: because I want to get used  to the best practice

Y-Havryliv (approved):
Well done!

PR #160, opened 2022-10-05
Solution

Mowka7:


app/test_main.py
```diff
@@ -1,3 +1,35 @@
...
+            id="should return [0, 0] for cat and dog with age 0"
+        ),
+        pytest.param(
+            14, 14, [0, 0],
+            id="should return [0, 0] for cat and dog with age < 15"
+        ),
+        pytest.param(
+            24, 23, [2, 1],
+            id="should return [2, 1] for cat and dog with age 24 and 23"
+        ),
```
TPolina, line 19: you also need to test 15, 23 and 24 years for both cats and dogs because these values are edge cases

TPolina (changes requested):

app/main.py
```diff
@@ -1,10 +1,15 @@
-def get_human_age(cat_age, dog_age):
+def get_human_age(cat_age: int, dog_age: int) -> list:
```
MasterpieceElbow, line 1: Resolve conflicts to run the tests

MasterpieceElbow (changes requested):

PR #161, opened 2022-10-05
Solution

YevheniiDavydenko:


TPolina (approved):
good, but better to add id to each test case

PR #169, opened 2022-10-05
Solution

LiudmylaKulinchenko:


mltdll (changes requested):
I'd suggest to add more tests. You lack at least two:
- test for larger ages, 
- test for the age where the same cat_age and dog_age no longer output the same result.

mltdll (changes requested):
You still don't have the test for the boundary where cat_years and dog_years diverge.

LiudmylaKulinchenko:
Або я тебе неправильно розумію, поясни тоді українською, будь ласка

LiudmylaKulinchenko:
Дякую, що пояснив 🐇

TPolina (approved):
Good job!

PR #173, opened 2022-10-05
Solution

fsocie7y:


rostyak (approved):

app/test_main.py
```diff
@@ -1,3 +1,40 @@
...
+            id="animal age less than 15"
+        ),
+        pytest.param(
+            16, 19, [1, 1],
+            id="animal age is more 15 and less 23"
+        ),
+        pytest.param(
+            23.9, 23.1, [1, 1],
+            id="animal age with remainder"
+        ),
```
TPolina, line 24: you need to test 14, 15, 23, 24 years for both cats and dogs, because these values are edge cases

TPolina (changes requested):

PR #178, opened 2022-10-08
Solution

Bohdan2001007:


app/test_main.py
```diff
@@ -1,3 +1,33 @@
 from app.main import get_human_age
 
-# write your code here
+
+def if_ages_are_zero() -> None:
```
IvanRamyk, line 4: test name should have `test_` prefix
```suggestion
def test_ages_are_zero() -> None:
```

IvanRamyk: fix everywhere

IvanRamyk (changes requested):

IvanRamyk (approved):
Well done!

PR #180, opened 2022-10-11
'Solution'

IvanGLS:


app/test_main.py
```diff
@@ -1,3 +1,6 @@
 from app.main import get_human_age
 
-# write your code here
+
+# Brevity is the soul of wit
```
IvanRamyk, line 4: Please, remove the extra comment :)

app/test_main.py
```diff
@@ -1,3 +1,6 @@
...
+def test_check_all_counted_parameters_cat_and_dog() -> None:
```
IvanRamyk, line 5: you should check all possible boundary values and equivalence partitions

IvanRamyk: https://www.guru99.com/equivalence-partitioning-boundary-value-analysis.html

IvanRamyk (changes requested):

app/test_main.py
IvanRamyk: What about `(14, 14)`
`(15, 15)`

IvanRamyk (changes requested):

IvanRamyk (approved):
Good job!

PR #184, opened 2022-10-13
Solution of py-cat-and-dog-years

YLobunets:


app/test_main.py
```diff
@@ -1,3 +1,45 @@
...
+    def test_should_convert_before_15_year_to_0(self) -> None:
+        assert get_human_age(5, 10) == [0, 0], \
+            "1-14 years should convert into 0 human age."
+
+    def test_should_convert_15_year_to_1(self) -> None:
+        assert get_human_age(15, 15) == [1, 1], \
+            "15 years should convert into 1 human age."
+
+    def test_should_convert_between_15_and_24_year_to_1(self) -> None:
+        assert get_human_age(18, 23) == [1, 1], \
```
DmytroSvirsa, line 28: better to test edged variant 23 for both: cat and dog

app/test_main.py
```diff
@@ -1,3 +1,45 @@
...
+            "from 15 up to 24 years should convert into 1 human age."
+
+    def test_should_convert_24_year_to_2(self) -> None:
+        assert get_human_age(24, 24) == [2, 2], \
+            "24 years should convert into 2 human age."
+
+    def test_should_convert_over_24_year_without_rounding(self) -> None:
+        assert get_human_age(28, 29) == [3, 3], \
+            "28/29 years should convert into 3 human age."
```
DmytroSvirsa, line 37: also will be good if you test the edged variant for 27/28

app/test_main.py
```diff
@@ -1,3 +1,45 @@
...
+        assert get_human_age(-5, -15) == [0, 0], \
+            "negative years should convert into 0 human age."
+
+    def test_should_convert_0_to_0(self) -> None:
+        assert get_human_age(0, 0) == [0, 0], \
+            "0 years should convert into 0 human age."
+
+    def test_should_convert_before_15_year_to_0(self) -> None:
+        assert get_human_age(5, 10) == [0, 0], \
+            "1-14 years should convert into 0 human age."
```
DmytroSvirsa, line 21: this one is not necessary if you test all edged  variants 0 and 14, so add a test for 14/14

DmytroSvirsa (changes requested):

app/test_main.py
YLobunets: Ok, corrected

app/test_main.py
YLobunets: Ok, variant 5/10 corrected to 14/14

app/test_main.py
YLobunets: Was added

DmytroSvirsa (approved):

PR #185, opened 2022-10-15
Solution

Mowka7:


app/test_main.py
```diff
@@ -1,3 +1,47 @@
...
+            id="should return [21, 17] for cat and dog with age 100"
+        )
+    ]
+)
+def test_get_human_age(
+        cat_age: int,
+        dog_age: int,
+        expected_array: list
+) -> None:
+    assert get_human_age(cat_age, dog_age) == expected_array
```
dmytro-petliovanyi, line 47: Is this code prints message with assertion error? cause maybe i just dont understand ¯\_(ツ)_/¯. like: f"Should return {human_age}, when cat age is {cat_years} and dog age is {dog_years}"

dmytro-petliovanyi (changes requested):

app/test_main.py
Mowka7: Yes, it is. I don't understand the reason of rejection

dmytro-petliovanyi: So if it is, i'm approving

dmytro-petliovanyi (approved):

TPolina (approved):
Good job!

PR #187, opened 2022-10-18
'Solution'

rakamakaphone:


app/test_main.py
```diff
@@ -1,3 +1,19 @@
...
+    assert get_human_age(28, 28) == [3, 2]
+
+
+def test_when_cat_age_is_less_than_0() -> None:
+    assert get_human_age(-10, 100) == [0, 17]
+
+
+def test_when_dog_age_is_less_than_0() -> None:
+    assert get_human_age(100, -1) == [21, 0]
+
```
DmytroSvirsa, line 15: test all edged variants, look for example in readme

DmytroSvirsa (changes requested):

DmytroSvirsa (changes requested):
what about 14, 14 result is 0,0 and 15,15 result 1,1 please make tests that will test critical points of function

rakamakaphone:
> what about 14, 14 result is 0,0 and 15,15 result 1,1 please make tests that will test critical points of function



> what about 14, 14 result is 0,0 and 15,15 result 1,1 please make tests that will test critical points of function

I thought i need to check critical values that we give to the function.

DmytroSvirsa (changes requested):
Yes, we need to check critical values, that we give to function, look precisely and find all of them

DmytroSvirsa (approved):

PR #188, opened 2022-10-19
Solution

amirammina:


app/test_main.py
```diff
@@ -1,3 +1,25 @@
 from app.main import get_human_age
 
-# write your code here
+
+def test_should_return_0_if_animal_age_less_then_15() -> None:
+    assert get_human_age(12, 12) == [0, 0]
```
DmytroSvirsa, line 5: better to test critical points like 14 here

app/test_main.py
```diff
@@ -1,3 +1,25 @@
...
+
+
+def test_should_return_1_if_animal_age_equal_to_15() -> None:
+    assert get_human_age(15, 15) == [1, 1]
+
+
+def test_should_return_1_if_animal_age_is_less_then_24() -> None:
+    assert get_human_age(20, 20) == [1, 1]
```
DmytroSvirsa, line 13: here the same 23 is better to test

DmytroSvirsa (changes requested):

DmytroSvirsa (approved):

PR #191, opened 2022-10-25
Solution

nailia-mamiedova:


Abnormaltype (changes requested):
add more tests. Check the checklist

nailia-mamiedova:
added tests with 0, negative number, large number, and test with raises TypeError when function receives an incorrect type of data

app/test_main.py
```diff
@@ -1,3 +1,55 @@
+import pytest
+
 from app.main import get_human_age
 
-# write your code here
+
+class TestConvertAge:
```
Abnormaltype, line 6: if you use pytest, you should write in function-based

Abnormaltype (changes requested):

Abnormaltype (approved):
Well done!

PR #207, opened 2022-11-23
Solution

mish-sk:


Abnormaltype (changes requested):
use checklist

app/test_main.py
```diff
@@ -1,3 +1,30 @@
+import pytest
+
 from app.main import get_human_age
 
-# write your code here
+
+def test_get_human_age() -> None:
```
Serhii-Leonenko, line 6: It's not a good descriptive  name for test

Serhii-Leonenko (changes requested):
Add more tests for all use cases, like an example test that 15 cat and dog years gives 1 year, but values lower than 15 like 14 - no, etc
image

app/test_main.py
```diff
@@ -1,3 +1,42 @@
...
+
+
+def test_animals_age_equals_to_two_human_years() -> None:
+    assert get_human_age(25, 27) == [2, 2]
+
+
+def test_if_output_changed_with_previous_value() -> None:
+    result_1 = get_human_age(30, 33)
+    result_2 = get_human_age(30, 33)
+    assert result_1 is not result_2
```
OleksandrPozden, line 25: You can write in simpler in one row

app/test_main.py
```diff
@@ -1,3 +1,40 @@
...
+def test_if_function_gives_correct_output_when_age_is_bigger_than_3() -> None:
+    assert get_human_age(88, 88) == [18, 14]
+
+
+def test_animals_age_equals_to_two_human_years() -> None:
+    assert get_human_age(25, 27) == [2, 2]
+
+
+def test_if_output_changed_with_previous_value() -> None:
+    assert get_human_age(30, 33) is not get_human_age(30, 33)
```
IvanRamyk, line 23: don't understand what do you want to check here

IvanRamyk (changes requested):

OleksandrPozden (approved):

PR #211, opened 2022-11-29
Solution

ochernyi:


app/test_main.py
```diff
@@ -1,3 +1,67 @@
...
+        ),
+        pytest.param(
+            24, 24, [2, 2],
+            id="result_should_be_correct"
+        ),
+        pytest.param(
+            100, 100, [21, 17],
+            id="result_should_be_correct"
+        ),
+        pytest.param(
```
IvanRamyk, line 21: I think it's better  to move this test to the beginning and put cases in the sort order

app/test_main.py
```diff
@@ -1,3 +1,67 @@
...
+            23, 23, [1, 1],
+            id="result_should_be_correct"
+        )
+    ]
+)
+def test_should_check_correct_data(
+        testing_cat_age: int,
```
IvanRamyk, line 32:
```suggestion
        cat_age: int,
```
I believe it's an extra prefix

IvanRamyk (changes requested):

OleksandrPozden (approved):

PR #214, opened 2022-12-04
Solution

Soobig666:


Serhii-Leonenko (changes requested):
Check the checklist. I don't see tests for all cases:
image

PR #220, opened 2022-12-07
Solution

viktory-koroliova:


app/test_main.py
```diff
@@ -1,3 +1,57 @@
+import pytest
 from app.main import get_human_age
+from typing import Any
 
-# write your code here
+
+class TestCatDogAge:
```
Abnormaltype, line 6: If u use pytest, use function base

Abnormaltype (changes requested):

Abnormaltype (approved):

PR #228, opened 2022-12-22
Solution

sorrow03:


app/test_main.py
```diff
@@ -1,3 +1,19 @@
...
+        (get_human_age(15, 15), [1, 1]),
+        (get_human_age(23, 23), [1, 1]),
+        (get_human_age(24, 24), [2, 2]),
+        (get_human_age(27, 27), [2, 2]),
+        (get_human_age(28, 28), [3, 2]),
+        (get_human_age(100, 100), [21, 17])
+    ]
+)
+def test_get_human_age(example_input: list, expected: list) -> None:
+    assert example_input == expected
```
katevasilenko, line 19: Seems like you can avoid some self repeating. Just write only arguments in parametrize and put them to a function in assert

katevasilenko (changes requested):

app/test_main.py
```diff
@@ -1,3 +1,19 @@
...
+        ((15, 15), [1, 1]),
+        ((23, 23), [1, 1]),
+        ((24, 24), [2, 2]),
+        ((27, 27), [2, 2]),
+        ((28, 28), [3, 2]),
+        ((100, 100), [21, 17])
+    ]
+)
+def test_get_human_age(example_input: tuple, expected: list) -> None:
+    assert get_human_age(example_input[0], example_input[1]) == expected
```
katevasilenko, line 19: Better, but still not clear enough. You can provide those `example_input` As two different variables and named them informative

katevasilenko (changes requested):

katevasilenko (approved):
Nice!

PR #230, opened 2022-12-25
Solution

evgenijmartynuk07:


app/test_main.py
```diff
@@ -1,3 +1,13 @@
 from app.main import get_human_age
 
-# write your code here
+
+def test_should_be_add_zero_when_age_0() -> None:
```
IvanRamyk, line 4: please, test more values from different classes (e.x for ages 2, 3, and more) and different boundary values (14-15), (23-24)

IvanRamyk (changes requested):

IvanRamyk (approved):

PR #235, opened 2023-01-02
Solution

ViacheslavYasynetskyi:


Abnormaltype (changes requested):
use the checklist before pushing your solution

app/test_main.py
```diff
@@ -1,3 +1,66 @@
...
+            )
+        ]
+    )
+    def test_should_raise_type_error_if_age_is_string(
+            self,
+            cat_age: Any,
+            dog_age: Any
+    ) -> None:
+        with pytest.raises(TypeError):
+            get_human_age(cat_age, dog_age)
```
olena-hul, line 66: We don't need parametrization if there is only a one case. Add more cases with different types, if you already use parametrization

app/test_main.py
```diff
@@ -1,3 +1,66 @@
...
+            pytest.param(
+                28, 29,
+                [3, 3],
+                id="28/29 cat/dog years should convert into 3 human age."
+            )
+        ]
+    )
+    def test_correct_output(
+            self,
+            cat_age: Any,
```
olena-hul, line 45: Add concrete annotation, don't use `Any`

olena-hul:
Add tests with floats and big numbers

olena-hul (changes requested):

PR #242, opened 2023-01-05
Solution

Tania-Kharchuk:


app/test_main.py
```diff
@@ -1,3 +1,86 @@
...
+            )
+        ]
+    )
+    def test_return_correct_values(
+            self,
+            input_cat_years: int,
+            input_dog_years: int,
+            expected_human_years: list
+    ) -> None:
+        assert get_human_age(input_cat_years, input_dog_years) \
```
katevasilenko, line 71: Can you change code to avoid \ please?

app/test_main.py
```diff
@@ -1,3 +1,86 @@
+import pytest
 from app.main import get_human_age
 
-# write your code here
+
+class TestGetHumanAge:
```
katevasilenko, line 5: Mmm… I’m not sure you need class to group tests as you have tests for only one function in this particular project. in the future, maybe. But not here. I would prefer to delete

katevasilenko (changes requested):

Abnormaltype (approved):
Well done!

PR #247, opened 2023-01-12
Solution

6aPa6aH:


app/test_main.py
```diff
@@ -1,3 +1,26 @@
 from app.main import get_human_age
 
-# write your code here
+import pytest
+
+
+class TestCatDogYears:
```
olena-hul, line 6: Add tests for negative numbers and incorrect types

olena-hul (changes requested):

app/test_main.py
```diff
@@ -1,3 +1,60 @@
...
+        ]
+    )
+    def test_negative_values_errors(
+            self,
+            cat_years: int,
+            dog_years: int,
+    ) -> None:
+        with pytest.raises(AssertionError):
+            assert cat_years > 0
+            assert dog_years > 0
```
ArsenPidhoretskyi, line 42: This test does not cover any option. You need to call function get_human_age

ArsenPidhoretskyi (changes requested):

olena-hul (approved):

olena-hul:
Good job!

PR #248, opened 2023-01-12
solution

AnatoliyPilipey:


app/test_main.py
```diff
@@ -1,3 +1,27 @@
 from app.main import get_human_age
+import pytest
 
-# write your code here
+
+def test_years_negative_or_equal_zero() -> None:
+    assert min(get_human_age(-1, 0)) == 0
```
Serhii-Leonenko, line 6: You should check that both negative value and zero give zero in the list result. So assert it with list [0. 0], because min() can work not correct

app/test_main.py
```diff
@@ -1,3 +1,27 @@
...
+    assert get_human_age(28, 28) == [3, 2]
+
+
+def test_long_life() -> None:
+    assert get_human_age(100, 100) == [21, 17]
+
+
+def test_receives_an_incorrect_type() -> None:
+    with pytest.raises(TypeError):
+        get_human_age("12", [12])
```
Serhii-Leonenko, line 27: You should add more tests,  for example, you check that 14 gives 0, but you don't check that 15 gives 1, need to check all boundaries. Also in all other similar cases.

Serhii-Leonenko (changes requested):

OleksandrPozden (approved):

PR #262, opened 2023-02-01
'Solution'

Paul-Maslov:


app/test_main.py
```diff
@@ -1,3 +1,88 @@
...
+            each_year: int,
+    ) -> None:
+
+        with pytest.raises(ZeroDivisionError):
+            convert_to_human(
+                initial_animal_age,
+                first_year,
+                second_year,
+                each_year
+            )
```
Bohdan-Salamakha, line 88: add tests on:
really big numbers,
negative numbers,
incorrect types of data

Bohdan-Salamakha (changes requested):
v1

app/test_main.py
Paul-Maslov: In principle, the additional tests made are redundant, since the data type check is carried out in the function, and the input data check must be done (if required by the customer) inside the function)

Paul-Maslov: Sorry my flake8 didn't notice mistake. Fixed

andrewbelo (approved):

Bohdan-Salamakha (approved):

PR #264, opened 2023-02-02
'Solution3'

timyr4iq:


app/test_main.py
```diff
@@ -1,3 +1,24 @@
...
+        (15, 15, [1, 1]),
+        (23, 23, [1, 1]),
+        (24, 24, [2, 2]),
+        (27, 28, [2, 2]),
+        (28, 29, [3, 3]),
+        (100, 100, [21, 17]),
+        (2000, 2000, [496, 397]),
+    ]
+)
+def test_cryptocurrency_action_result(
```
katevasilenko, line 19: The name of test is not represent what you are testing

katevasilenko (changes requested):
Please add test with incorrect type of data

ArsenPidhoretskyi (approved):
good job!

PR #269, opened 2023-02-08
solution

VladyslavOladko:


app/test_main.py
```diff
@@ -1,3 +1,33 @@
 from app.main import get_human_age
 
-# write your code here
+
+def test_should_return_zero() -> None:
+    cat_age = 12
+    dog_age = 12
+    actually_result = get_human_age(cat_age=cat_age, dog_age=dog_age)
+    expected_result = [0, 0]
+    assert actually_result == expected_result
```
Vasyl-Smutok, line 9: on my mind there is no need to create additional variables here

```suggestion
    assert get_human_age(12, 12) == [0, 0]
```

Vasyl-Smutok (changes requested):

PR #274, opened 2023-02-13
Solution

oleksandr-lysanets:


ArsenPidhoretskyi (changes requested):
Add test when send string(must raise)
Also with big number will be good)

ArsenPidhoretskyi (changes requested):
Well done!)

ArsenPidhoretskyi (approved):

PR #281, opened 2023-02-15
Solution

AlimFreight:


ArsenPidhoretskyi (changes requested):
Use pytest.parametrize. Also add checks for string types and negative numbers

Bohdan-Salamakha (approved):

PR #282, opened 2023-02-16
Solution

MaybeTI:


app/test_main.py
```diff
@@ -1,3 +1,34 @@
...
+        (0, 0, [0, 0]),
+        (-27, 15, [0, 1]),
+        (14, 14, [0, 0]),
+        (15, 15, [1, 1]),
+        (23, 23, [1, 1]),
+        (24, 24, [2, 2]),
+        (27, 27, [2, 2]),
+        (28, 28, [3, 2]),
+        (100, 100, [21, 17])
+    ]
```
DavidGlusch, line 18: i think you are suposed to have description of each test (see Testing/Pytest parametrize video,  on platform, start at 10:30)

DavidGlusch (changes requested):

app/test_main.py
```diff
@@ -15,6 +15,17 @@
...
+    ids=[
+        "should return correct result",
+        "should return correct result",
+        "should return correct result",
+        "should return correct result",
+        "should return correct result",
+        "should return correct result",
+        "should return correct result",
+        "should return correct result",
+        "should return correct result"
```
NEkropo1, line 28: Mb a little bit more descriptive ids names?)
Can't get fast which test testes which case, just try to make it according to best cases)

app/test_main.py
```diff
@@ -15,6 +15,17 @@
...
     ]
 )
 def test_correct_result(cat: int, dog: int, result: list) -> None:
```
NEkropo1, line 31: and here you can try to rename func to something like: `test_cat_dog_age_should_equal_humans`

NEkropo1 (changes requested):

NEkropo1 (changes requested):
Well, I can't get why you make it more complicated. But anyway you finally provided some description for test cases. 
Your code now is not bad, still it lack few things, according to readme.md:

NEkropo1 (approved):
Seems ok for this exact case, mb someone could provide more information )
A more descriptive test_func name prob

Serhii-Leonenko (changes requested):

YuriiKindrat (approved):

PR #284, opened 2023-02-17
Solution

DavidGlusch:


app/test_main.py
```diff
@@ -1,3 +1,60 @@
+import pytest
+
 from app.main import get_human_age
 
-# write your code here
+
+class TestGetHumanAge:
+    @pytest.mark.parametrize(
+        "cat_age,dog_age,human_years",
```
WDemidenko, line 8: You can replace it with one tuple of ages, and then unpack it inside your function in test as arguments

WDemidenko (approved):
Good job!

app/test_main.py
```diff
@@ -1,3 +1,60 @@
...
+
+    @pytest.mark.parametrize(
+        "cat_age,dog_age,expected_error",
+        [
+            pytest.param(
+                "28", "28", TypeError,
+                id="expected to raise TypeError when not int or float passed"
+            )
+        ]
+    )
```
YuriiKindrat, line 52: no need to use mark.parametrize for just one case, use simple test for this

YuriiKindrat (changes requested):

app/test_main.py
```diff
@@ -1,3 +1,52 @@
...
+            dog_age: int,
+            human_years: list
+    ) -> None:
+
+        if human_years == TypeError:
+            with pytest.raises(TypeError):
+                get_human_age(cat_age, dog_age)
+
+        else:
+            assert get_human_age(cat_age, dog_age) == human_years
```
Serhii-Leonenko, line 52: You should split the test which checks the data type, but as Yurii said, you can make it without pytest.mark.parametrize

Serhii-Leonenko (changes requested):

app/test_main.py
```diff
@@ -1,3 +1,47 @@
...
+            pytest.param(
+                1256, 1256, [310, 248],
+                id="should work properly with big numbers"
+            ),
+            pytest.param(
+                25.2, 25.2, [2.0, 2.0],
+                id="should work properly with float values"
+            )
+        ]
+    )
```
Serhii-Leonenko, line 35: You should add more tests for all age boundaries, for example, you check here that 14, 14 gives 0, 0 in the result, but you don't check that 15, 15 gives 1, 1 in the result, etc.

image

Serhii-Leonenko (changes requested):

YuriiKindrat (approved):

PR #287, opened 2023-02-18
Solution type(False) is not int

NEkropo1:


SafonovVladimir (approved):
GJ

app/test_main.py
```diff
@@ -1,3 +1,50 @@
...
+            dog_years: int | float,
+            human_years: int | float
+    ) -> None:
+        assert get_human_age(cat_years, dog_years) == human_years
+
+
+class TestErrorTypes:
+    def test_type_error_missing_1_positional_argument(self) -> None:
+        with pytest.raises(TypeError, match=r".* missing 1 *."):
+            get_human_age(1)
```
Serhii-Leonenko, line 46: Interesting, but I think it's a redundant test because you will get the error in any case when there is a missing required argument, no need to check this case)

Serhii-Leonenko (changes requested):

app/test_main.py
NEkropo1: I agree, but all those tests are redundant since we already got type annotations for arguments in the defined `app.main.get_human_age()`
So If someone who uses code provides data not according to sample, it's his logical error, not program...

Still, according to checklist, I should provide :
`if the function receives an incorrect type of data it raises the correct exception.`
which, for me, is the same redundant test, so I added some other of same kind, just in case

That's why I will change code only if checklist will be changed, your suggestions seems redundant to me)

Serhii-Leonenko: Tests are such a thing that they can be written endlessly if we are guided by the logic that we need to test every case. But there should be a reasonable limit to which cases you should test. If the QA wants to annoy the developer, he will make a lot of tests such like this ))))

Serhii-Leonenko (approved):
Well done!

PR #289, opened 2023-02-19
solution

usrofgh:


DmytryTong (approved):
Well.. Using classes is interesring way to solve this, but I think you can solve the task without them (second class of TypeError's testing could be in only one function). But, it looks more complex with the classes, so I think - it'll work this way too. Good job)

app/test_main.py
```diff
@@ -1,3 +1,53 @@
...
+        ],
+    )
+    def test_raising_error_correctly(
+            self,
+            cat_age: Any,
+            dog_age: Any,
+            expected_error: [TypeError]
+    ) -> None:
+        with pytest.raises(expected_error):
+            get_human_age(cat_age, dog_age)
```
Serhii-Leonenko, line 53: Don't create an extra class for this test, you can move it to TestConvertAnimalAgeToHuman class

app/test_main.py
```diff
@@ -1,3 +1,53 @@
...
+    @pytest.mark.parametrize(
+        "cat_age,dog_age,expected_ages",
+        [
+            (0, 0, [0, 0]),
+            (14, 14, [0, 0]),
+            (15, 15, [1, 1]),
+            (23, 23, [1, 1]),
+            (24, 24, [2, 2]),
+            (27, 27, [2, 2]),
+            (28, 28, [3, 2]),
```
Serhii-Leonenko, line 18: What about 29 for dog?) You can use not similar age in these both tuples.

app/test_main.py
```diff
@@ -1,3 +1,53 @@
...
+            (100, 100, [21, 17]),
+            (-1, -1, [0, 0]),
+        ],
+    )
```
Serhii-Leonenko, line 22: Also add descriptive ids to your tests

Serhii-Leonenko (changes requested):

YuriiKindrat (changes requested):
fix flake8 errors

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #291, opened 2023-02-20
Solution

ZabFTFT:


Morty67 (approved):

app/test_main.py
```diff
@@ -1,3 +1,72 @@
...
+                23,
+                [1, 1],
+                id="should calculate second year"
+            ),
+            pytest.param(
+                28,
+                28,
+                [3, 2],
+                id="should calculate third year differently"
+            ),
```
Serhii-Leonenko, line 34: what about 24 for both or 27 for a cat and 29 for a dog?) You should check all age boundaries

app/test_main.py
```diff
@@ -1,3 +1,72 @@
...
+                -100,
+                -100,
+                [0, 0],
+                id="should return 0 if input age is negative"
+            ),
+            pytest.param(
+                0,
+                20,
+                [0, 1],
+                id="should return 0 if input age is zero"
```
Serhii-Leonenko, line 57: it's better to use both zero values

Serhii-Leonenko (changes requested):

kuznetsb (approved):

YuriiKindrat (approved):

PR #292, opened 2023-02-20
Solution DonchykD

DmytryTong:


SergiiMachulin (approved):
LGTM, but i suggest to add negative numbers check according to the checklist of README.md

app/test_main.py
```diff
@@ -1,3 +1,56 @@
...
+            100,
+            [21, 17],
+            id="test should return result with large arguments"
+        )
+    ]
+)
+def test_of_get_human_age(
+        cat_age: int,
+        dog_age: int,
+        expected_result: list) -> None:
```
Serhii-Leonenko, line 50: fix indentation

Serhii-Leonenko (changes requested):
Agree with the previous comment, add a check for negative numbers, also I don't see tests for all age boundaries, what about it:

image

app/test_main.py
DmytryTong: Fixed: negative number case and 27/27, 28/28 cases added. Identation - fixed.

YuriiKindrat (approved):

PR #293, opened 2023-02-20
'Solution'

Morty67:


Maytreyaya (approved):

app/test_main.py
```diff
@@ -1,3 +1,63 @@
...
+                0,
+                [0, 0],
+                id="Human age is zero when pets age are zero",
+            ),
+            pytest.param(
+                14,
+                14,
+                [0, 0],
+                id="Human age is zero when pets age < 15"
+            ),
```
Serhii-Leonenko, line 20: You check here that 14 gives  0 in the result, but do you check that  15 gives  1 in the result?)
add more tests to check all age boundaries

Serhii-Leonenko (changes requested):

app/test_main.py
```diff
@@ -1,3 +1,69 @@
+import pytest
 from app.main import get_human_age
```

Serhii-Leonenko (approved):

PR #294, opened 2023-02-20
CatDog Solution

Maytreyaya:


amber-marichi (approved):
very nice code, but there is one moment: shouldn't we make local libraries import after standard and related libraries according to import rules?

app/test_main.py
```diff
@@ -1,3 +1,69 @@
...
+            pytest.param(1000, 1000, [246, 197],
+                         id="test working with big numbers"
+                         )
+        ]
+    )
+    def test_age(self,
+                 cat_age: int,
+                 dog_age: int,
+                 expected: list
+                 ) -> None:
```
Serhii-Leonenko, line 50: fix indentation here

app/test_main.py
```diff
@@ -1,3 +1,69 @@
...
+                id="different human age of dog and cat at 28 years"
+            ),
+            pytest.param(
+                45, 45, [7, 6],
+                id="different age of cat and dog with big age"
+            ),
+            pytest.param(1000, 1000, [246, 197],
+                         id="test working with big numbers"
+                         )
+        ]
```
Serhii-Leonenko, line 44: What about 27 for cat or 29 for dog, also 14 or 23 for both?)

app/test_main.py
```diff
@@ -1,3 +1,69 @@
+import pytest
 from app.main import get_human_age
```

Serhii-Leonenko (changes requested):

app/test_main.py
```diff
@@ -1,3 +1,80 @@
...
+    @pytest.mark.parametrize(
+        "cat_age,dog_age,expected_error",
+        [
+            pytest.param(
+                "dog", 4, TypeError,
+                id="not int was entered"
+
+            )
+        ]
+    )
```
YuriiKindrat, line 74: no need to use decorator for just one case

YuriiKindrat (changes requested):

app/test_main.py
Maytreyaya: should i remove it or add new params?

Maytreyaya: okay, according to youe message i will delete it

NickSavchyn (approved):
good luck with mentors XD

app/test_main.py
```diff
@@ -1,3 +1,67 @@
...
+                28, 28, [3, 2],
+                id="different human age of dog and cat at 28 years"
+            ),
+            pytest.param(
+                45, 45, [7, 6],
+                id="different age of cat and dog with big age"
+            ),
+            pytest.param(1000, 1000, [246, 197],
+                         id="test working with big numbers"
+                         )
```
YuriiKindrat, line 56: fix indentation here

app/test_main.py
```diff
@@ -1,3 +1,67 @@
+import pytest
+
 from app.main import get_human_age
 
-# write your code here
+
+class TestGetHumanAge:
+    @pytest.mark.parametrize(
+
```
YuriiKindrat, line 8: delete unnecessary blank line here

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #295, opened 2023-02-20
Solution

SergiiMachulin:


BohdanLazaryshyn (approved):

app/test_main.py
```diff
@@ -1,3 +1,41 @@
...
+    "cat_age,dog_age,human_age",
+    [
+        (-1, -1, [0, 0]),
+        (0, 0, [0, 0]),
+        (14, 14, [0, 0]),
+        (15, 15, [1, 1]),
+        (23, 23, [1, 1]),
+        (24, 24, [2, 2]),
+        (27, 27, [2, 2]),
+        (28, 28, [3, 2]),
```
Serhii-Leonenko, line 21: Looks good, but I advise you to make some changes to these data, in the first you can use both 27, 28 values,  in the second also you can use both values which will give 3 in the result

Serhii-Leonenko (changes requested):

Serhii-Leonenko (approved):
Good job!

PR #299, opened 2023-02-21
Solution

yura-hudzovskyi:


NazarHladaniuk (approved):

app/test_main.py
```diff
@@ -1,3 +1,51 @@
...
+            pytest.param(
+                28, 28, [3, 2],
+                id="should return [3, 2]"
+            ),
+            pytest.param(
+                100, 100, [21, 17],
+                id="should return [21, 17]"
+            )
+        ]
+    )
```
YuriiKindrat, line 39: add test for negtive numbers, also it is better to add f.e. "when ages is equal to [<ur test cases>]" to ids

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #301, opened 2023-02-21
Solution

oksana-feshchenko:


kuznetsb (approved):
I think you should also add test for raising TypeError if we pass not integer to get_human_age()

YuriiKindrat (changes requested):
add tests for this, please
image

wQuelS (approved):
Nice

PR #303, opened 2023-02-21
Solution

viktoria-rybenchuk:


ada-krav (approved):

app/test_main.py
```diff
@@ -1,3 +1,27 @@
...
+def test_should_return_correct_age_for_the_next_24_years() -> None:
+    assert get_human_age(24, 24) == [2, 2]
+
+
+def test_should_return_correct_age_for_the_next_cat_years() -> None:
+    assert get_human_age(28, 28) == [3, 2]
+
+
+def test_should_return_correct_age_for_the_next_dog_years() -> None:
+    assert get_human_age(29, 29) == [3, 3]
```
YuriiKindrat, line 22: use mark.parametrize for similar tests
also make more tests, what about big and negative numbers?

YuriiKindrat (changes requested):

wQuelS (approved):
Nice

PR #304, opened 2023-02-21
Solution

NazarHladaniuk:


app/test_main.py
```diff
@@ -1,3 +1,17 @@
+import pytest
+
```
destr34, line 2: good code, but you dont need an empty line when you import modules

destr34 (changes requested):

app/test_main.py
NazarHladaniuk: This is true only for modules of the same type. Between third-party imports and local application specific imports, you should have an empty line.

Danil228 (approved):

app/test_main.py
```diff
@@ -1,3 +1,17 @@
...
 from app.main import get_human_age
 
-# write your code here
+
+@pytest.mark.parametrize("age1, age2, expected", [
```
YuriiKindrat, line 6: age1, age2 is not descriptive names for variable

YuriiKindrat (changes requested):
add tests for this cases
image

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #305, opened 2023-02-21
Solution

ada-krav:


MikhailLyvak (approved):
All is good, but realy shure that mentors will propose u use DRY here)

app/test_main.py
```diff
@@ -1,3 +1,28 @@
...
+def test_1_human_year() -> None:
+    assert get_human_age(23, 23) == [1, 1]
+
+
+def test_2_human_years() -> None:
+    assert get_human_age(24, 25) == [2, 2]
+
+
+def test_more_human_years() -> None:
+    assert get_human_age(29, 35) == [3, 4]
```
YuriiKindrat, line 28: use mark.parametrize for this similar tests
also make tests for more cases, f.e. what about big numbers?

YuriiKindrat (changes requested):

wQuelS (approved):
Nice

PR #306, opened 2023-02-21
Solution

destr34:


RuslanLopatin (approved):

app/test_main.py
```diff
@@ -1,3 +1,33 @@
...
+def test_animals_27_years_old() -> None:
+    assert get_human_age(27, 27) == [2, 2]
+
+
+def test_animals_28_years_old() -> None:
+    assert get_human_age(28, 28) == [3, 2]
+
+
+def test_animals_100_years_old() -> None:
+    assert get_human_age(100, 100) == [21, 17]
```
YuriiKindrat, line 33: use mark.parametrize for similar tests
also add more test cases, check ur checklist

YuriiKindrat (changes requested):

app/test_main.py
```diff
@@ -1,33 +1,57 @@
...
+        ),
+        (
+            -2, -25, [0, 0]
+        )
+    ]
+)
+def test_negative_age(cat_age: int,
+                      dog_age: int,
+                      expected_years: list) -> None:
+    assert get_human_age(cat_age, dog_age) == expected_years
```
destr34, line 57: i didn't add test for float numbers, because we have type annotation

wQuelS: yeah, but type annotations do not secure your function from passing wrong arguments. You should test behavior of your function if wrong data type is passing to your function

wQuelS (changes requested):

Serhii-Leonenko (changes requested):
All checks should be passed

app/test_main.py
```diff
@@ -1,3 +1,86 @@
...
+        ),
+        (
+            100.3, 100.9, [21, 17]
+        )
+    ]
+)
+def test_float_numbers_age(cat_age: int,
+                           dog_age: int,
+                           expected_years: list) -> None:
+    assert get_human_age(cat_age, dog_age) == expected_years
```
YuriiKindrat, line 86: no need to create separate tests for floats

app/test_main.py
```diff
@@ -1,3 +1,86 @@
...
+            -1, 8, [0, 0]
+        ),
+        (
+            2, -10, [0, 0]
+        ),
+        (
+            -2, -25, [0, 0]
+        )
+    ]
+)
```
YuriiKindrat, line 53: add ids to ur tests

YuriiKindrat (changes requested):

wQuelS (changes requested):
Why do u still separate tests for negative numbers?
What if, for example, `str` is passed to your function? Make a test for such case

app/test_main.py
```diff
@@ -1,3 +1,86 @@
...
+        pytest.param(
+            28.1, 28.7, [3, 2],
+            id="should return 3, 2"
+        ),
+        pytest.param(
+            100.3, 100.9, [21, 17],
+            id="should return 21, 17"
+        )
+    ]
+)
```
Serhii-Leonenko, line 77: You still have test duplicates. No need to add 3 tests for negative numbers, add only one for this case. The same in other cases, such as float numbers. Remove redundant tests and better add the check that 29 for a dog returns 3 in the result.

app/test_main.py
```diff
@@ -1,3 +1,86 @@
+import pytest
 from app.main import get_human_age
```
Serhii-Leonenko, line 2: You should split different groups of imports with a blank line

Serhii-Leonenko (changes requested):

app/test_main.py
```diff
@@ -1,3 +1,87 @@
...
+        pytest.param(
+            27.1, 27.2, [2, 2],
+            id="should return 2, 2"
+        ),
+        pytest.param(
+            28.1, 28.7, [3, 2],
+            id="should return 3, 2"
+        ),
+        pytest.param(
+            100.3, 100.9, [21, 17],
```
wQuelS, line 74: so many similar checks..

wQuelS (changes requested):

YuriiKindrat (approved):

PR #309, opened 2023-02-22
'Solution'

NickSavchyn:


oksana0410 (approved):

app/test_main.py
```diff
@@ -1,3 +1,21 @@
...
+def test_first_15_years():
+    assert get_human_age(15, 23) == [1, 1]
+
+
+def test_next_9_years():
+    assert get_human_age(25, 25) == [2, 2]
+
+
+def test_every_next_5_years():
+    assert get_human_age(28, 28) == [3, 2]
```
YuriiKindrat, line 21: use mark.parametrize for similar tests
also test more cases, check a checklist of a task

YuriiKindrat (changes requested):

app/test_main.py
```diff
@@ -1,3 +1,42 @@
...
 
-# write your code here
+
+@pytest.mark.parametrize(
+    "cat_age, dog_age, human_age",
+    [
+        (-1, -1, [0, 0]),
+        (0, 0, [0, 0]),
+        (14, 14, [0, 0]),
+        (15, 23, [1, 1]),
```
Serhii-Leonenko, line 12: in this case you check one only for a dog and the other one only for a cat, but you should check them for both

app/test_main.py
```diff
@@ -1,3 +1,42 @@
...
+        (25, 25, [2, 2]),
+        (28, 28, [3, 2]),
```
Serhii-Leonenko, line 14: what about 29 for a dog?)

app/test_main.py
```diff
@@ -1,3 +1,42 @@
...
+
+@pytest.mark.parametrize("cat_age, dog_age, human_age", [
+    ("one", 1, TypeError),
+    (1, "one", TypeError),
+    ("one", "one", TypeError),
+])
+def test_get_human_age_errors(
+        cat_age: int,
+        dog_age: int,
+        human_age: Any
```
Serhii-Leonenko, line 39: are you sure that Any?)

app/test_main.py
```diff
@@ -1,3 +1,42 @@
+import pytest
+from typing import Any
 from app.main import get_human_age
```
Serhii-Leonenko, line 3: image

Serhii-Leonenko (changes requested):

Serhii-Leonenko (commented):
All checks should be passed

wQuelS (changes requested):
All tests should pass

app/test_main.py
Serhii-Leonenko: still not fixed

app/test_main.py
Serhii-Leonenko: still not fixed

Serhii-Leonenko (changes requested):

Serhii-Leonenko (approved):

PR #313, opened 2023-02-22
Solution

oigod:


AnnaBoichenko (approved):

app/test_main.py
```diff
@@ -1,3 +1,65 @@
...
+        )
+    ]
+)
+def test_get_human_age_correct(
+        cat_age: int,
+        dog_age: int,
+        expected: list
+) -> None:
+    assert get_human_age(cat_age, dog_age) == expected
+    # veeeery hard task(((
```
YuriiKindrat, line 64: what about testing negative numbers ?)
also remove ur comment, please )

YuriiKindrat (changes requested):

app/test_main.py
YuriiKindrat: also fix ur flake8 errors

YuriiKindrat (changes requested):
all tests should pass, fix ur flake8 errors
image

YuriiKindrat (approved):

PR #315, opened 2023-02-22
Solution

VladyslavCherkashyn:


app/test_main.py
```diff
@@ -1,3 +1,47 @@
 from app.main import get_human_age
 
-# write your code here
+import pytest
+
+
+class TestGetHumanAge:
+    @pytest.mark.parametrize(
+        "cat_age, dog_age, expected_res", [
```
AlessaQ, line 8: Better expected_result

app/test_main.py
```diff
@@ -1,3 +1,47 @@
...
+            pytest.param(
+                14,
+                14,
+                [0, 0],
+                id="under_lower_boundary_condition"
```
AlessaQ, line 13: The description does not need to use lower space`_`

app/test_main.py
```diff
@@ -1,3 +1,47 @@
...
+            ),
+            pytest.param(
+                15,
+                23,
+                [1, 1],
+                id="condition_for_return_1"
```
AlessaQ, line 19: The description should describe tests, your description is more appropriate for the title.

AlessaQ (changes requested):

app/test_main.py
```diff
@@ -1,3 +1,47 @@
...
+        "cat_age, dog_age, expected_result", [
+            pytest.param(
+                14,
+                14,
+                [0, 0],
+                id="under lower boundary condition"
+            ),
+            pytest.param(
+                15,
+                23,
```
Serhii-Leonenko, line 17: In this case, you test one value only for the cat but not for the dog, the same in other cases. You should check all age boundaries for both of them.

Serhii-Leonenko (changes requested):
Read the checklist!

image

wQuelS (approved):
Good

PR #321, opened 2023-02-22
Solution

ramsty:


me1nyk (changes requested):
I was at a QnA meeting and mentor mentioned that we need to use pytest.mark.parametrize to solve this task:
# Parametrization - is a feature of the Pytest framework that allows you to run the same test case with different parameters. So, pytest.mark.parameterize allows you to make testing more efficient and productive, simplifying the process of writing tests and reducing their duplication.
And in this task we have like same test case with different parameters.
Also don't forget to check if type of input data correct (you can find this comment in checklist : if the function receives an incorrect type of data it raises the correct exception.)

ramsty:
I see, no problem to rewrite this code, but there's nothing about it in documentation if I see right)

me1nyk (approved):

Serhii-Leonenko (changes requested):
Read the checklist, you should add more tests to check such cases as negative numbers, and raising error, also you need add tests for all age boundaries (15, 24 ...). And according to the previous comment we recommend using pytest.mark.parametrize in this task to simplify the code.

YuriiKindrat (approved):

PR #322, opened 2023-02-22
Solution

AlessaQ:


dogmoony (changes requested):
In checklist has been said that you need to check if TypeError is raised while using function. Add the TypeError test.

app/test_main.py
```diff
@@ -1,3 +1,34 @@
...
 from app.main import get_human_age
 
-# write your code here
+
+class TestGetHumanAge:
+    @pytest.mark.parametrize(
+        "cat_age,dog_age,expected",
+        [
+            (-1, -1, [0, 0]),
+            (14, 1, [0, 0]),
```
Serhii-Leonenko, line 11: in this case, you test one value only for the cat but not for the dog, the same in other cases. You should check all age boundaries for both of them

Serhii-Leonenko: and what about 29 for a dog?)

app/test_main.py
```diff
@@ -1,3 +1,34 @@
...
+    )
+    def test_get_human_age(
+        self, cat_age: int, dog_age: int, expected: list
+    ) -> None:
+        assert get_human_age(cat_age, dog_age) == expected
+
+
+def test_wrong_data_type_str() -> None:
+    with pytest.raises(TypeError):
+        get_human_age("1", 1)
```
Serhii-Leonenko, line 34: you can put this test inside the class

Serhii-Leonenko (changes requested):

app/test_main.py
AlessaQ: fixed

app/test_main.py
AlessaQ: fixed

Serhii-Leonenko (approved):
Good job!

PR #323, opened 2023-02-22
'Solution'

AnnaBoichenko:


VladyslavCherkashyn (approved):

app/test_main.py
```diff
@@ -1,3 +1,17 @@
...
+
+@pytest.mark.parametrize(
+    "cat_age,dog_age,result",
+    [
+        (0, 14, [0, 0]),
+        (15, 23, [1, 1]),
+        (28, 23, [3, 1]),
+        (15, 29, [1, 3])
+    ]
+)
```
Serhii-Leonenko, line 13: Add descriptive ids

app/test_main.py
```diff
@@ -1,3 +1,17 @@
+import pytest
 from app.main import get_human_age
```

Serhii-Leonenko (changes requested):
Read the checklist!
You should test all age boundaries and other cases

app/test_main.py
```diff
@@ -1,3 +1,20 @@
+import pytest
+
 from app.main import get_human_age
 
-# write your code here
+
+@pytest.mark.parametrize(
+    "cat_age,dog_age,result",
+    [
+        pytest.param(0, 14, [0, 0], id="animals' age is less then 15"),
```
Serhii-Leonenko, line 9: in this case, you test one value only for the cat but not for the dog, the same in other cases. You should check all age boundaries for both of them

Serhii-Leonenko (changes requested):
And I repeat you again READ THE CHECKLIST!

image

app/test_main.py
AnnaBoichenko: sorry, I don't get it
both 0 and 14 in one boundary range, less than 15. so it would be the same test if I run it for 0, 0 or 14, 14, or  3, 12 for instance

AnnaBoichenko: I also don`t understand, why I should test for an incorrect data type. In this case, the function doesn`t check for input data type and doesn't raise custom exceptions. So if a user ignores type annotations and passes a string to the function, it will raise TypeError: '<' not supported between instances of 'str' and 'int'
Am I supposed to check for negative numbers?
Any "large number" is also in the same rule range. So test for a test for 25 and 125 years is basically the same test

YuriiKindrat: @AnnaBoichenko u should check an incorrect data type to test if function works correctly with incorrect data type and raises exception when it needed, the same is for other cases, so implement all required tests in readme.md file
Imagine, that u don't know how this function works inside, so u should test all possible conditions for it

YuriiKindrat (changes requested):
Add more tests, for all conditions from ur readme.md file and checklist

app/test_main.py
```diff
@@ -1,3 +1,38 @@
...
+    assert get_human_age(cat_age, dog_age) == result
+
+
+def test_cannot_pass_other_types_to_function() -> None:
+    with pytest.raises(TypeError):
+        get_human_age("12", "_")
+    with pytest.raises(TypeError):
+        get_human_age(["12"], ["_"])
+    with pytest.raises(TypeError):
+        get_human_age({"12"}, {"_"})
```
YuriiKindrat, line 38: no need to use with for 3 times here, u can use parametrize or just call func for 3 times

app/test_main.py
```diff
@@ -1,3 +1,38 @@
...
+        pytest.param(
+            16, 0, [1, 0], id="cat's age is more than 15 less than 24"
+        ),
+        pytest.param(
+            0, 23, [0, 1], id="dog's age is more than 15 less than 24"
+        ),
+        pytest.param(24, 23, [2, 1], id="cat's age is 24"),
+        pytest.param(15, 24, [1, 2], id="dog's age is 24"),
+        pytest.param(28, 23, [3, 1], id="cat's age is more than 24"),
+        pytest.param(15, 29, [1, 3], id="dog's age is more than 24"),
```
YuriiKindrat, line 23: what about big/negative nums ?

YuriiKindrat (changes requested):

app/test_main.py
AnnaBoichenko: @YuriiKindrat why big numbers are different from any number more than 24?

AnnaBoichenko: > Add more tests, for all conditions from ur readme.md file and checklist

I'm literally trying to follow the very first rule in the checklist - Make sure that your tests don't test the same case few times.
i got it about why we have to check cat's and dog's age separately. because an implementation of the function might be specific for each animal.
negative nums are strange things to pass to the function, but I'll add the test. 
but it seems that 100 and 30 must work the same, using the same logic at the same "rule range"

PR #325, opened 2023-02-22
Solution

Litvinenko23:


app/test_main.py
```diff
@@ -1,3 +1,27 @@
...
+    pytest.param("0", 0, TypeError, id="cat age is not an integer"),
+])
+def test_get_human_age(cat_age: int,
+                       dog_age: int,
+                       expected: list | TypeError) -> None:
+    if isinstance(expected, list):
+        assert get_human_age(cat_age, dog_age) == expected
+    else:
+        with pytest.raises(TypeError):
+            get_human_age(cat_age, dog_age)
```
Zenionoi, line 27: You also can create another function to check for raises `TypeError` and avoid `if/else` statement

Zenionoi (approved):

app/test_main.py
Serhii-Leonenko: Agree, it's better to move this check to another function

app/test_main.py
```diff
@@ -1,3 +1,27 @@
...
+    pytest.param(14, 14, [0, 0], id="both pets are under 15 years old"),
+    pytest.param(15, 15, [1, 1], id="both pets are exactly 15 years old"),
+    pytest.param(23, 23, [1, 1], id="both pets are under 24 years old"),
+    pytest.param(24, 24, [2, 2], id="both pets are exactly 24 years old"),
+    pytest.param(27, 27, [2, 2], id="both pets are under 28 years old"),
+    pytest.param(28, 28, [3, 2], id="both pets are exactly 28 years old"),
+    pytest.param(100, 100, [21, 17], id="both pets are over 28 years old"),
+    pytest.param(-1, -1, [0, 0], id="pet has negative age"),
+    pytest.param("0", 0, TypeError, id="cat age is not an integer"),
+])
```
Serhii-Leonenko, line 19: What about 29 for a dog?

Serhii-Leonenko (changes requested):

PR #329, opened 2023-02-22
'Solution'

Bohdan-Kovalenko1992:


oksanaaam (approved):

app/test_main.py
```diff
@@ -1,3 +1,56 @@
...
+
+    @pytest.mark.parametrize(
+        "cat_age_input, dog_age_input, expected_error",
+        [
+            pytest.param(
+                "1", 1, TypeError,
+                id="type cat_age_input and dog_age_input is int"
+            )
+        ]
+    )
```
YuriiKindrat, line 48: no need to use parametrize for just 1 case

app/test_main.py
```diff
@@ -1,3 +1,56 @@
 from app.main import get_human_age
 
-# write your code here
+import pytest
+
```
YuriiKindrat, line 4: fix imports according to pep8

app/test_main.py
```diff
@@ -1,3 +1,56 @@
...
+            pytest.param(
+                100, 100, [21, 17],
+                id="should return [21, 17]"
+            ),
+            pytest.param(
+                -20, -20, [0, 0],
+                id="should return [0, 0]"
+            ),
+        ]
+    )
```
YuriiKindrat, line 31: add more cases
image
also, what about big numbers ?

YuriiKindrat (changes requested):

Y-Havryliv (approved):
GJ!

PR #331, opened 2023-02-22
solution

Yulia-Mala:


app/test_main.py
```diff
@@ -1,3 +1,28 @@
...
+
+    ]
+)
+def test_convert_animal_years_to_human(cat_years: int,
+                                       dog_years: int,
+                                       result: list) -> None:
+    assert get_human_age(cat_years, dog_years) == result, (
+        (f"{cat_years} cat years and {dog_years} dog years "
+         f"are equal to {result} human years")
+    )
```
ZhAlexR, line 28: Have a look to the requirements inside checklist:
1) you should have a test for the wrong type values like "str"
2) you should test negative values too
3) you should test work of function with laaaaaarge numbers))))
And check one more time that you have covered all edge cases

ZhAlexR (changes requested):

app/test_main.py
```diff
@@ -1,3 +1,36 @@
...
+@pytest.mark.parametrize(
+    "cat_years,dog_years,result",
+    [
+        pytest.param(15, 15, [1, 1], id="exactly one year"),
+        pytest.param(24, 24, [2, 2], id="exactly two years"),
+        pytest.param(28, 29, [3, 3], id="exactly three years"),
+        pytest.param(32, 34, [4, 4], id="exactly four years"),
+        pytest.param(0, 0, [0, 0], id="zero years"),
+        pytest.param(7, 14, [0, 0], id="zero years - we discard the reminder"),
+        pytest.param(15, 20, [1, 1], id="1 year, we discard the reminder"),
```
Serhii-Leonenko, line 14: image

No need to use the same values, but it's better to add checks for all age boundaries for both(cat and dog). For example, you check that 14 gives 0 in the result only for a dog, but what about a cat.  Also, make more descriptive ids

Serhii-Leonenko (changes requested):

wQuelS (approved):

PR #341, opened 2023-02-28
Solution

Anyta17:


ArsenPidhoretskyi (changes requested):
Add tests with big values
Add tests with incorrect type

Bohdan-Salamakha (approved):

PR #349, opened 2023-03-03
Solution

StreetNik:


ArsenPidhoretskyi (changes requested):
Add tests with big values(>=100) and so small(<14, <0)
Also add tests with wrong type

ArsenPidhoretskyi (approved):
Well done!

PR #353, opened 2023-03-07
Solution

Andrew13X:


ArsenPidhoretskyi (changes requested):
Use pytest parametrize
Add tests with values less than 0, more that 100, also values with wrong types

ArsenPidhoretskyi (approved):
Well done!

PR #358, opened 2023-03-10
'Solution'

IvanGuculyak:


Vasyl-Smutok (changes requested):
Add test with wrong type
image

Grayder0152 (approved):

PR #360, opened 2023-03-14
solution

DmytroProkofiev:


ArsenPidhoretskyi (changes requested):
Add tests with wrong types also with negative numbers

ArsenPidhoretskyi (approved):
Well done!

PR #361, opened 2023-03-14
'Solution'

KirillMelanich:


app/test_main.py
```diff
@@ -1,3 +1,50 @@
...
+def test_if_age_more_then_28_and_29_then_convert_to_3() -> None:
+    age = get_human_age(28, 29)
+
+    assert age == [3, 3]
+
+
+def test_if_the_age_is_negative() -> None:
+    age = get_human_age(-15, -15)
+
+    assert age == [0, 0]
```
Grayder0152, line 50: Use `parametrize` fixture for thore

Grayder0152 (changes requested):

vasylhnatiuk (changes requested):
do test if the function receives an incorrect type of data it raises the correct exception.

ArsenPidhoretskyi (approved):
Well done!

PR #368, opened 2023-03-30
'Solution'

StinGER-ua:


app/test_main.py
```diff
@@ -1,3 +1,33 @@
...
+def test_get_animal_age_27() -> None:
+    assert get_human_age(27, 27) == [2, 2]
+
+
+def test_get_animal_age_28() -> None:
+    assert get_human_age(28, 28) == [3, 2]
+
+
+def test_get_animal_age_100() -> None:
+    assert get_human_age(100, 100) == [21, 17]
```
ArsenPidhoretskyi, line 33: Use pytest.parametrize

ArsenPidhoretskyi (commented):
Add tests with wring types. And values lowest than 0

ArsenPidhoretskyi (changes requested):
Changes requested

app/test_main.py
```diff
@@ -1,3 +1,31 @@
...
+        (0, 0, [0, 0]),
+        (14, 14, [0, 0]),
+        (15, 15, [1, 1]),
+        (23, 23, [1, 1]),
+        (24, 24, [2, 2]),
+        (27, 27, [2, 2]),
+        (28, 28, [3, 2]),
+        (100, 100, [21, 17]),
+        ("14", 14, TypeError),
+        (14, "14", TypeError),
```
ArsenPidhoretskyi, line 19: Create another test with wrong types no need to write them all in one.
You need to have two tests:
1. With int values
2. With the wrong types

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #375, opened 2023-04-10
Tests done

dirolius:


app/test_main.py
Bohdan-Salamakha, line 1: add tests for normal values like in example
https://github.com/mate-academy/py-cat-and-dog-years#:~:text=for%20writing%20tests.-,Examples%3A,-get_human_age(0

Bohdan-Salamakha (changes requested):
v1

app/test_main.py
```diff
@@ -1,3 +1,66 @@
...
+                15,
+                15,
+                [1, 1],
+                id="does not work correctly with a normal value"
+            ),
+            pytest.param(
+                28,
+                28,
+                [3, 2],
+                id="does not work correctly with a normal value"
```
Bohdan-Salamakha, line 55: Why id="does not work correctly with a normal value" in good cases?
It works correctly

app/test_main.py
```diff
@@ -1,3 +1,66 @@
...
+                0,
+                0,
+                [0, 0],
+                id="does not work correctly with the value zero"
+            ),
+            pytest.param(
+                100,
+                100,
+                [21, 17],
+                id="does not work correctly with a large value"
```
Bohdan-Salamakha, line 19: it's a valid case too.

Creme Puff is the oldest cat ever at 38 years and 3 days old. She was born on August 3, 1967, and died on August 6, 2005.
Creme Puff was owned by Granpa Rex Allen's owner and started her day with bacon and eggs, asparagus, broccoli, and coffee with heavy cream each morning.

Bohdan-Salamakha (changes requested):
v2

ArsenPidhoretskyi (approved):
Well done!

PR #377, opened 2023-04-10
Solution

callogan:


app/test_main.py
```diff
@@ -1,3 +1,25 @@
 from app.main import get_human_age
 
-# write your code here
+
+def test_should_return_zero_when_input_zero_years() -> None:
+    assert get_human_age(0, 0) == [0, 0]
```
olena-hul, line 5: Add tests for negative values and for wrong types (str, list for example)

olena-hul (changes requested):

app/test_main.py
```diff
@@ -1,3 +1,37 @@
...
+
+def test_should_convert_extra_dog_years() -> None:
+    assert get_human_age(16, 29) == [1, 3]
+
+
+def test_incorrect_type_of_data() -> None:
+    try:
+        get_human_age("28", [28])
+    except TypeError:
+        assert "'<' not supported between instances of 'str' and 'int'"
```
Bohdan-Salamakha, line 37: assert does nothing
use
`raise AssertionError("Operations with these types aren't yet supported")`

Bohdan-Salamakha (changes requested):
v1

ArsenPidhoretskyi (changes requested):
Use pytest.parametrize due to code duplicate

Bohdan-Salamakha (approved):

PR #380, opened 2023-04-12
Solution

goldenuni:


app/test_main.py
Bohdan-Salamakha, line 1: add tests for the negative numbers, invalid types and really big numbers

Bohdan-Salamakha (changes requested):
v1

app/test_main.py
goldenuni: @Bohdan-Salamakha, what do it mean really big numbers?

Bohdan-Salamakha (approved):

PR #383, opened 2023-04-19
Solution cat and dog years

vborovets0:


app/test_main.py
```diff
@@ -1,3 +1,32 @@
+import pytest
 from app.main import get_human_age
 
-# write your code here
+
+class TestMain:
```
olena-hul, line 5: Fix flake8

olena-hul (changes requested):

Bohdan-Salamakha (changes requested):
add test for correct input type

EdAlekseiev (approved):
Good job

PR #386, opened 2023-04-25
Solution

Fawnygod:


app/test_main.py
```diff
@@ -1,3 +1,17 @@
...
+def test_two_human_years() -> None:
+    assert get_human_age(27, 28) == [2, 2]
+
+
+def test_three_human_years() -> None:
+    assert get_human_age(29, 29) == [3, 3]
+
+
+def test_extra_human_years() -> None:
+    assert get_human_age(50, 50) == [8, 7]
```
ArsenPidhoretskyi, line 17: Use pytest.paramatrize. 
Add tests with big values, with values < 0 also values in range(0; 15), with wrong type

ArsenPidhoretskyi (changes requested):

PR #391, opened 2023-05-03
problem with file not found (python 3.11)

chipapapo:


app/test_main.py
```diff
@@ -1,3 +1,37 @@
 from app.main import get_human_age
 
-# write your code here
+# fifteen is the age of both cat and dog turn 1 year old
+# twenty-four is the age of both cat and dog turn 2 years old
+# twenty-eight is the age of cat turns 3 years old and dog still 2 years old
```
Grayder0152, line 5: You can move those comments

Grayder0152 (approved):

PR #395, opened 2023-05-08
Solution

OlehOryshchuk:


app/test_main.py
```diff
@@ -1,3 +1,33 @@
+import pytest
+
 from app.main import get_human_age
 
-# write your code here
+
+@pytest.mark.parametrize(
+    ("cat_age", "dog_age", "expected_human_age"),
+    [
+        pytest.param(14, 14, [0, 0],
```
olena-hul, line 9: Add tests for negatives and for big numbers

olena-hul (changes requested):

PR #398, opened 2023-05-11
Solution

VladyslavKaminskyi:


app/test_main.py
```diff
@@ -1,3 +1,87 @@
...
+            34,
+            "4",
+            TypeError,
+            id="Raise TypeError if age is not integer",
+        )
+    ],
+)
+def test_raising_errors_correctly(cat_age: int,
+                                  dog_age: int,
+                                  expected_error: Any) -> None:
```
ArsenPidhoretskyi, line 85: Wrong typing must be
type[Exception] or Type[Exception] (Type is from typing module)

PR #405, opened 2023-05-20
Py cat and dog years

alexgubarev88:


app/test_main.py
```diff
@@ -1,3 +1,54 @@
...
+            ),
+            pytest.param(
+                100,
+                100,
+                [21, 17],
+                id="should return [21, 17] if both ages are equal 100"
+            )
+        ]
+    )
+    def test_correctly_convert_animal_age(
```
ArsenPidhoretskyi, line 47: Add tests with wrong types

ArsenPidhoretskyi: Also with values < 0

ArsenPidhoretskyi (changes requested):

Vasyl-Smutok (approved):
Good Job 👍

PR #414, opened 2023-05-24
Develop

contritum:


Bohdan-Salamakha (changes requested):
add tests for both invalid input and output types

contritum:
> add tests for both invalid input and output types

But in the lecture "Test best practices" at 3:10 lector said that creating tests for annotation types is a bad practice.

olena-hul:
> > add tests for both invalid input and output types
> 
> But in the lecture "Test best practices" at 3:10 lector said that creating tests for annotation types is a bad practice.

According to a checklist: 'if the function receives an incorrect type of data it raises the correct exception.'
Generally, it is a bad practice, it's True, but sometimes we need to ensure that the correct exception is risen in the function if some condition occurrs. In this task, we should follow the checklist and add tests for wrong types

olena-hul (changes requested):
Please fix comments

contritum:
> Please fix comments

Which comments? What's wrong?

olena-hul:
> > Please fix comments
> 
> Which comments? What's wrong?

I meant the above comments about adding tests for wrong types. It is done already, ok

olena-hul (approved):

PR #417, opened 2023-05-26
solution

sbilikepy:


vsmutok (changes requested):
add tests for incorrect types

sbilikepy:
> add tests for incorrect types
```
def test_numeric_values_in_returned_list() -> None:
    for age in get_human_age(15, 15.0):
        assert isinstance(age, (int, float))
```
should i change it? If yes - вo I need to add a function argument check for all types except `int` and `float`?

YukihiroSM (changes requested):
It's not really about handling -> that means, that the function throws an exception. Try to check, that the function really throws an exception in case of wrong types of input data. 
Try takin a look into with pytest.raises

sbilikepy:
image

Bohdan-Salamakha (approved):

sbilikepy:
Approve it again if solution is correct please
image

Abnormaltype (approved):

PR #444, opened 2023-06-13
Solution

Gvilion:


app/test_main.py
YevheniiIovchu, line 1: You skipped flake8 check, so make it.
And read carefully checklist. You should make check for exceptions and you can also make check that function cannot take negative numbers for age.

YevheniiIovchu (changes requested):

app/test_main.py
YevheniiIovchu, line 1: You still missed separete test with exception of incorrect type of put data when either list or str is put as arg into function and should call TypeError

YevheniiIovchu (changes requested):

app/test_main.py
YevheniiIovchu: You have flake8 error again. All the rest is ok.

YevheniiIovchu (changes requested):
You have flake8 error again. All the rest is ok.

YevheniiIovchu (approved):

app/test_main.py
```diff
@@ -1,3 +1,76 @@
...
+    "cat_age,dog_age",
+    [
+        pytest.param(
+            [10, 10], 100,
+            id="cat and dog age should be integer"
+        ),
+        pytest.param(
+            "100", 100,
+            id="dog and cat ages should be integer"
+        )
```
YuriiKindrat, line 69: what about dog age, will this test check it ?

YuriiKindrat (changes requested):

app/test_main.py
```diff
@@ -1,3 +1,84 @@
...
+            100, "100",
+            id="dog and cat ages should be integer"
+        )
+    ]
+)
+def test_get_human_age_with_wrong_data_type(
+        cat_age: int, dog_age: int
+) -> None:
+    with pytest.raises(TypeError):
+        get_human_age(cat_age, dog_age)
```
Oleg-s-s, line 84: What if function will not raise a TypeError?

Oleg-s-s (changes requested):

Oleg-s-s (approved):
Good job

PR #447, opened 2023-06-13
Develop

YevheniiIovchu:


app/test_main.py
```diff
@@ -1,3 +1,69 @@
...
+
+@pytest.mark.parametrize(
+    "cat_age,dog_age",
+    [
+        pytest.param(0, 0, id="Function argument cannot be negative")
+    ]
+)
+def test_get_human_age_with_negative_values(cat_age: int,
+                                            dog_age: int) -> None:
+    assert cat_age >= 0 and dog_age >= 0
```
Oleg-s-s, line 34: You can make better test with actual using of negative values and add it to next decorator so there will be no need in using additional one

Oleg-s-s (changes requested):

PR #448, opened 2023-06-14
solution

oleksiikolii:


app/test_main.py
```diff
@@ -1,3 +1,17 @@
 from app.main import get_human_age
 
-# write your code here
+
+def test_zero_when_age_less_14() -> None:
+    assert get_human_age(14, 14) == [0, 0]
+
```
kovaliskoveronika, line 6: Instead of writing a few tests, you can use @pytest.mark.parametrize to transfer parameters into a function. 
https://docs.pytest.org/en/7.2.x/how-to/parametrize.html
And also maybe check how the function will work with unusual parameters( for ex negative numbers or zero)

kovaliskoveronika (changes requested):

kovaliskoveronika (approved):

app/test_main.py
```diff
@@ -1,3 +1,58 @@
...
+    @pytest.mark.parametrize("cat,dog,expected", [
+        (28, 28, [3, 2]),
+        (100, 100, [21, 17])
+    ])
+    def test_age_dog_and_cat_should_be_different(self,
+                                                 cat: int,
+                                                 dog: int,
+                                                 expected: list) -> None:
+        assert get_human_age(cat, dog) == expected
```
YuriiKindrat, line 46: it seems to me u can write only one test for this with ids usage )

YuriiKindrat (changes requested):

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #449, opened 2023-06-14
Solution

Tarasidze:


app/test_main.py
```diff
@@ -1,3 +1,49 @@
...
+    )
+    def test_get_human_age(self,
+                           cat_age: int,
+                           dog_age: int,
+                           exp_result: list) -> None:
+        assert (
+            get_human_age(cat_age, dog_age) == exp_result
+        ), f"Cat:{cat_age} and dog:{cat_age} should return {exp_result}"
+
+    @pytest.mark.parametrize(
```
yakhlamova, line 35: That`s not a mistake, but you can make a separate class for raising error.

app/test_main.py
```diff
@@ -1,3 +1,49 @@
...
+
+
 from app.main import get_human_age
 
-# write your code here
+
+class TestCatAndDogYears:
+    @pytest.mark.parametrize(
+        "cat_age, dog_age, exp_result",
+        [
```
yakhlamova, line 10: your tests don`t check negative numbers situations, please add

yakhlamova (changes requested):

app/test_main.py
```diff
@@ -1,3 +1,51 @@
...
+
+    @pytest.mark.parametrize(
+        "cat_age, dog_age, expected_error",
+        [
+            pytest.param("",
+                         17,
+                         TypeError,
+                         id="Should return error when type is incorrect")
+        ]
+    )
```
YuriiKindrat, line 45: no need to use parametrize for only one case, u can write simple test for this, or add more cases )

YuriiKindrat (changes requested):

app/test_main.py
Tarasidze: I've thought about checking the second parameter by incorrect type, and forgot)

YuriiKindrat (approved):

PR #450, opened 2023-06-14
Solution

DmytroTsehelnyk:


YKonoplyov (approved):
Good job👍

oleksiikolii (approved):
Hi, I also got your task to review, nvm I'm just approving to not get in trouble later :)

YuriiKindrat (changes requested):
image
what about this test?)

app/test_main.py
```diff
@@ -1,3 +1,54 @@
...
+            )
+        ]
+    )
+    def test_raise_correct_exception(
+            self,
+            cat_age: any,
+            dog_age: any,
+    ) -> None:
+        with pytest.raises(TypeError):
+            get_human_age(cat_age, dog_age)
```
YuriiKindrat, line 54: why do u need to use parametrize for only one case ?
also what if dog years won't raise an typeerror ?

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #453, opened 2023-06-14
My tests

kovaliskoveronika:


app/test_main.py
```diff
@@ -1,3 +1,34 @@
+import pytest
+
 from app.main import get_human_age
 
-# write your code here
+
+@pytest.mark.parametrize("value,result",
+                         [((14, 14), [0, 0]),
```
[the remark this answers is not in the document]
Altahoma: However, I think your solution is also workable. _Summoning menthors for help._

app/test_main.py
```diff
@@ -1,3 +1,34 @@
...
+@pytest.mark.parametrize("value",
+                         [(1, "hh"),
+                          (1, []),
+                          (2, {}),
+                          (1, ())])
+def test_wrong_type(value: tuple) -> None:
+    try:
+        get_human_age(*value)
+    except TypeError:
+        pass
```
YuriiKindrat, line 34: u can use pytest.raises here

YuriiKindrat: also it is better to add ids to all of ur tests

YuriiKindrat (changes requested):

app/test_main.py
```diff
@@ -1,3 +1,58 @@
...
+                         ids=[
+                             "wrong type string",
+                             "wrong type list",
+                             "wrong type dict",
+                             "wrong type set"
+                         ])
+def test_wrong_type(cat_age: int, dog_age: any) -> None:
+    if not isinstance(dog_age, int):
+        with pytest.raises(TypeError):
+            get_human_age(cat_age, dog_age)
```
Oleg-s-s, line 58: That's wrong approach, you should better make smth like in `test_usual_case` with expected result

app/test_main.py
```diff
@@ -1,3 +1,60 @@
...
+                                  expected_result: list) -> None:
+    assert get_human_age(cat_age, dog_age) == [0, 0]
+
+
+@pytest.mark.parametrize("cat_age, dog_age, error",
+                         [
+                             (1, "hh", TypeError),
+                             (1, [], TypeError),
+                             (2, {}, TypeError),
+                             (1, (), TypeError)
```
LiudmylaKulinchenko, line 49: What if cat_age will have a wrong type, but dog_age won't?

app/test_main.py
```diff
@@ -1,3 +1,60 @@
...
+                         ],
+                         ids=[
+                             "age equal zeros",
+                             "age is negative number",
+                             "one of age is float",
+                             "one of age is bool"
+                         ])
+def test_data_out_of_normal_range(cat_age: int,
+                                  dog_age: int,
+                                  expected_result: list) -> None:
```
LiudmylaKulinchenko, line 40: Why do you need expected_result variable here? You don't use it in the test
So just move these test cases in the previous test, it will work the same and you will not need this one

LiudmylaKulinchenko (changes requested):

PR #456, opened 2023-06-14
Cat and dog years(first commit):

IhorVoskoboinikov:


app/test_main.py
```diff
@@ -1,3 +1,24 @@
...
+        pytest.param(0, 15, [0, 1], id="cat below 15, dog age to 15"),
+        pytest.param(10**10, 10**20, [2499999996, 19999999999999999997],
+                     id="big numbers"),
+        pytest.param(25, -50, [2, 0], id="negative dog age"),
+        pytest.param(-2, 35, [0, 4], id="negative cat age"),
+        pytest.param(-2, -1, [0, 0], id="both negative age"),
+    ],
+)
+def test_get_human_age(cat_age: int, dog_age: int, expected: int) -> None:
+    assert get_human_age(cat_age, dog_age) == expected
```
Andriy-Sydorenko, line 24: You can add an additional test function for the incorrect type of input data, e.g. if one of the arguments has `str` type.

app/test_main.py
```diff
@@ -1,3 +1,24 @@
...
+        pytest.param(28, 28, [3, 2], id="cat age greater than dog age"),
+        pytest.param(100, 100, [21, 17], id="both ages greater than 28"),
+        pytest.param(15, 0, [1, 0], id="cat equal to 15, dog below 15"),
+        pytest.param(0, 15, [0, 1], id="cat below 15, dog age to 15"),
+        pytest.param(10**10, 10**20, [2499999996, 19999999999999999997],
+                     id="big numbers"),
+        pytest.param(25, -50, [2, 0], id="negative dog age"),
+        pytest.param(-2, 35, [0, 4], id="negative cat age"),
+        pytest.param(-2, -1, [0, 0], id="both negative age"),
+    ],
```
Andriy-Sydorenko, line 21: In my opinion, you should write more descriptive ids, so it will be clear what exactly went wrong and why

Andriy-Sydorenko (changes requested):

PR #457, opened 2023-06-14
Implemented tests for convert_to_human()

Altahoma:


app/test_main.py
```diff
@@ -1,3 +1,48 @@
+import pytest
+from typing import Any
+
 from app.main import get_human_age
 
-# write your code here
+
+class TestGetHumanAge:
```
volodymyr-komarnyckyi, line 7: No need to use class here

volodymyr-komarnyckyi (changes requested):

app/test_main.py
Altahoma: Using classes allows to logically group related test cases together.

volodymyr-komarnyckyi: We only have tests for get_human_age, so I thought there's no reason to add class
But let it be
Overall good job

volodymyr-komarnyckyi (approved):

app/test_main.py
```diff
@@ -1,3 +1,48 @@
...
+            pytest.param(
+                24, 28, [2, 2],
+                id="Should be 2 when: 23 < cat age < 28 or 23 < dog age < 29",
+            ),
+            pytest.param(
+                28, 29, [3, 3],
+                id="Should increase by 1 every 4 or 5 for cat and dog over 24",
+            )
+        ]
+    )
```
YuriiKindrat, line 28: add more tests, did u read checklist ?

app/test_main.py
```diff
@@ -1,3 +1,48 @@
...
+        [
+            (str()),
+            (list()),
+            (set()),
+            (tuple()),
+        ]
+    )
+    def test_raising_error(self, cat_age: Any) -> None:
+        with pytest.raises(TypeError):
+            get_human_age(cat_age, 0)
```
YuriiKindrat, line 48: what if typeerror won't be raised for dog age ? :)

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #474, opened 2023-06-25
Solution

NazarLytvyniuk:


app/test_main.py
```diff
@@ -1,3 +1,26 @@
...
+    [
+        (0, 0, [0, 0]),
+        (14, 14, [0, 0]),
+        (15, 15, [1, 1]),
+        (23, 23, [1, 1]),
+        (24, 24, [2, 2]),
+        (27, 27, [2, 2]),
+        (28, 28, [3, 2]),
+        (29, 29, [3, 3]),
+        (100, 100, [21, 17])
```
Dimosphen1, line 17: Consider adding test cases to cover input with negative values, different values for cat_age and dog_age, and to test that the correct exception is raised in case of incorrect data type input.

app/test_main.py
```diff
@@ -1,3 +1,26 @@
...
+    ]
+)
+def test_should_return_correct_ages(
+        cat_age: int,
+        dog_age: int,
+        expected_human_ages: list[int]
+) -> None:
+
```
Dimosphen1, line 25: Exceptions processing may include changes in the `test_should_return_correct_ages` function as well.

Dimosphen1 (changes requested):
Additional test cases requested

app/test_main.py
```diff
@@ -1,3 +1,49 @@
...
+    assert get_human_age(cat_age, dog_age) == expected_human_ages
+
+
+@pytest.mark.parametrize(
+    "cat_age, dog_age",
+    [
+        ([], 12),
```
Dimosphen1, line 36: Parametrize should include the expected value as well. For this test it would be `exception_type`, rewrite this test to use the following signature "cat_age,dog_age,exception_type".

app/test_main.py
```diff
@@ -1,3 +1,49 @@
...
+        (6, []),
+        (5, "dog")
+    ]
+)
+def test_should_raise_type_error(
+        cat_age: Any,
+        dog_age: Any
+) -> None:
+
+    with pytest.raises(Exception) as error:
```
Dimosphen1, line 47: Test that test raises exactly the exception that is needed, not a general exception.

app/test_main.py
```diff
@@ -1,3 +1,49 @@
...
+        get_human_age(cat_age, dog_age)
+    assert error.type == TypeError
```
Dimosphen1, line 49: Avoid using `assert` for this test, all work can be done in a context manager.

Dimosphen1 (changes requested):
Several changes to `test_should_raise_type_error` were requested.

Polyakiv-Andrey (approved):
Good job!

PR #475, opened 2023-06-25
Solution cat and dog years

MarkShapka:


app/test_main.py
```diff
@@ -1,3 +1,58 @@
...
+        ),
+        pytest.param(
+            27, 27, [2, 2], id="test_input_ages_equal_27"
+        ),
+        pytest.param(
+            28, 28, [3, 2], id="test_input_ages_equal_28"
+        ),
+        pytest.param(
+            29, 29, [3, 3], id="test_input_ages_equal_29"
+        )
```
Dimosphen1, line 34: Add test cases to check for negative input, large numbers, and different values for cat_age and dog_age.

app/test_main.py
```diff
@@ -1,3 +1,58 @@
...
+    ]
+)
+def test_ages(cat_age: int, dog_age: int, expect_result: list) -> None:
+    actual = get_human_age(cat_age, dog_age)
+    assert actual == expect_result
```
Dimosphen1, line 39: Avoid using the `actual` variable here, there is no need to allocate the variable only for assertion when you can make it with function directly.

Dimosphen1 (changes requested):
Several changes needed

app/test_main.py
```diff
@@ -1,3 +1,36 @@
...
+        (0, 0, [0, 0]),
+        (1, 1, [0, 0]),
+        (14, 14, [0, 0]),
+        (15, 15, [1, 1]),
+        (23, 23, [1, 1]),
+        (24, 24, [2, 2]),
+        (27, 27, [2, 2]),
+        (28, 28, [3, 2]),
+        (29, 29, [3, 3]),
+        (150, 150, [33, 27])
```
Dimosphen1, line 18: Requested tests added except the one that covers different values for `cat_age` and `dog_age`. Encounter this test as well (e.g. passing 28 and 15, 14 and 29 to the `get_human_age` function to test that produced output is correct).

app/test_main.py
```diff
@@ -1,3 +1,36 @@
...
+    ]
+)
+def test_ages(cat_age: int, dog_age: int, expect_result: list) -> None:
+    assert get_human_age(cat_age, dog_age) == expect_result
+
+
+@pytest.mark.parametrize(
+    "cat_age,dog_age,expected_error",
+    [
+        pytest.param("15", 15, TypeError, id="should be raise TypeError"),
```
Dimosphen1, line 28: Do not use `pytest.param` here as well, there is no added value from two similar ids.

Dimosphen1 (changes requested):
Several minor changes were requested regarding test cases and the use of `pytest.param`.

Dimosphen1 (approved):
Well done, nice work!

PR #479, opened 2023-06-28
Test branch

stepband:


vasylhnatiuk (changes requested):
image
add tests for incorect data

vasylhnatiuk (changes requested):
You must add a separate test for exceptions and test if the function already raises the correct exception with the wrong data.

ArsenPidhoretskyi (approved):
Well done!

PR #495, opened 2023-07-18
Solution

FadieievaA:


app/test_main.py
```diff
@@ -1,3 +1,25 @@
...
+        pytest.param(100, 100, [21, 17])
+    ]
+)
+def test_human_years_values(
+        cat_years: int,
+        dog_years: int,
+        human_years: list[int]
+) -> None:
+    assert (get_human_age(cat_years, dog_years) == human_years),\
+        f"Sum of {cat_years} and {dog_years} should be equal to {human_years}"
```
YukihiroSM, line 25: Don't use `\` for line separations. Better use `(` `)`

YukihiroSM (changes requested):
Looks good, but fix an issue in the comment below

app/test_main.py
FadieievaA: if i add ( ) then I got an error
app\test_main.py:24:5: F631 assertion is always true, perhaps remove parentheses?

Bohdan-Salamakha (changes requested):
add tests for negative values and incorrect types

PR #496, opened 2023-07-19
Solved

Dimo-n:


app/test_main.py
```diff
@@ -1,3 +1,91 @@
+import pytest
 from app.main import get_human_age
 
-# write your code here
+
+class TestGetHumanAge:
```
pavlejviki, line 5: using classes for tests is not mandatory in pytest. You can write standalone test functions

pavlejviki (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #497, opened 2023-07-19
Implemented tests

lubart2517:


app/test_main.py
```diff
@@ -1,3 +1,21 @@
 from app.main import get_human_age
+import pytest
 
-# write your code here
+
+@pytest.mark.parametrize(
+    "cat_age,dog_age,result",
+    [
+        pytest.param(0, 0, [0, 0], id="get_from_zeroes"),
```
Dimosphen1, line 8: Do not use `pytest.param` here, there is no need for that. Use just tuple in list, eg `[(0, 0, [0, 0]), (14, 14, [0, 0]), ...]`

app/test_main.py
```diff
@@ -1,3 +1,21 @@
...
+        pytest.param(24, 24, [2, 2], id="get_two_from_smallest"),
+        pytest.param(27, 27, [2, 2], id="get_two_from_biggest"),
+        pytest.param(28, 27, [3, 2], id="get_three_for_cat"),
+        pytest.param(100, 100, [21, 17], id="test_hundreds")
+    ]
+)
+def test_get_result(cat_age: int, dog_age: int, result: list[str]) -> None:
+    assert (
+        get_human_age(cat_age, dog_age) == result
+    ), f"Get human age of {cat_age} and {dog_age} should be equal to {result}"
```
Dimosphen1, line 21: There is no need in the second f string as well, leave `assert get_human_age(cat_age, dog_age) == result` only.

app/test_main.py
```diff
@@ -1,3 +1,21 @@
...
+    "cat_age,dog_age,result",
+    [
+        pytest.param(0, 0, [0, 0], id="get_from_zeroes"),
+        pytest.param(14, 14, [0, 0], id="get_zeroes_from_positive"),
+        pytest.param(15, 15, [1, 1], id="get_ones"),
+        pytest.param(23, 23, [1, 1], id="get_ones_from_the_biggest"),
+        pytest.param(24, 24, [2, 2], id="get_two_from_smallest"),
+        pytest.param(27, 27, [2, 2], id="get_two_from_biggest"),
+        pytest.param(28, 27, [3, 2], id="get_three_for_cat"),
+        pytest.param(100, 100, [21, 17], id="test_hundreds")
```
Dimosphen1, line 15: Add additional test cases to test against negative input and incorrect types of data. In case an incorrect type is passed, the function should raise the corresponding exception. Test that the raised exception is right.

Dimosphen1 (changes requested):
Additional test cases were requested according [to the checklist](https://github.com/mate-academy/py-cat-and-dog-years/blob/master/checklist.md) with several changes in code.

Polyakiv-Andrey (approved):
Good job!

PR #508, opened 2023-07-26
v1

YegorT:


YukihiroSM (changes requested):
Tests look good. But can you, please, add one more test which will test the function work with the examples like in description and use @pytest.mark.parametrize. 
Also, take a look into checklist: https://github.com/mate-academy/py-cat-and-dog-years/blob/master/checklist.md

Add test for exception raising in case of wrong input.

YukihiroSM (approved):
Looks good to me.

PR #514, opened 2023-08-03
Solution

ip-ilya:


app/test_main.py
```diff
@@ -1,3 +1,32 @@
...
+        28, 28, [3, 2],
+        id="should return [3, 2] when both ages are 29"
+    ),
+])
+def test_get_human_age_good(
+        cat_age: int,
+        dog_age: int,
+        expected_result: list
+) -> None:
+    assert get_human_age(cat_age, dog_age) == expected_result
```
pavlejviki, line 32: Add tests for negative numbers, really large numbers and that the function raises the correct exception when  it receives an incorrect type of data.

pavlejviki (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #533, opened 2023-08-22
Added tests

RPodolets:


app/test_main.py
```diff
@@ -1,3 +1,54 @@
...
+                14,
+                [0, 0],
+                id="test cat and dog age below first year"
+            ),
+            pytest.param(
+                15,
+                15,
+                [1, 1],
+                id="test cat and dog age equal to first year"
+            ),
```
denys-source, line 21: I would also add a test that checks the case when dog/cat age is 23

katryana, line 21: It would be better to add a check for 23 to determine the boundary

katryana (changes requested):
Good job

PR #534, opened 2023-08-23
Solution for py-cat-and-dog-years

Barsh4ec:


KristianYe (changes requested):
the checklist says that you should check the age type: 
image
so maybe you will get the request changes(
and also for negative values check:
image

PR #536, opened 2023-08-23
solution

innyshka:


app/test_main.py
```diff
@@ -1,3 +1,46 @@
+import pytest
+
 from app.main import get_human_age
 
-# write your code here
+
+class TestGetHumanAge:
```
OlegatorLE, line 6: why did you use class here?

app/test_main.py
```diff
@@ -1,3 +1,46 @@
...
+                0, 0, [0, 0], id="zero age should give list with zeros"
+            ),
+            pytest.param(
+                1200, 1200, [296, 237], id="big ages"
+            ),
+            pytest.param(
+                14, 14, [0, 0], id="pet's age is less than 1 human's year"
+            ),
+            pytest.param(
+                15, 15, [1, 1], id="15 pet years equals 1 human year"
```
OlegatorLE, line 23: this test and next one did the same?

app/test_main.py
```diff
@@ -1,3 +1,46 @@
...
+    )
+    def test_get_human_age(
+            self,
+            cat_age: int,
+            dog_age: int,
+            expected_result: list
+    ) -> None:
+        assert get_human_age(cat_age, dog_age) == expected_result
+
+    def test_correct_param(self) -> None:
```
OlegatorLE, line 44: I think you can extend these tests with @pytest.mark.parameterize

OlegatorLE (changes requested):
I think you can do better

PR #538, opened 2023-08-24
Solution

denys-source:


ORENYT (approved):
I will approve, but I'm not shure that your code is right, 
image
checklist says that you have to fix it, but video said that we don't have to test correct data input if we implemented correct types of variables

app/test_main.py
```diff
@@ -1,3 +1,57 @@
...
+            [1, 1],
+            id="23 dog/cat years should convert to 1 human year",
+        ),
+        pytest.param(
+            24,
+            24,
+            [2, 2],
+            id="24 dog/cat years should convert to 2 human years",
+        ),
+        pytest.param(
```
IvanRamyk, line 38: I'd add a test with big numbers (up to 100) to test that after 28 boundary everything works fine

IvanRamyk: It's also mentioned in the checklist

IvanRamyk (changes requested):

PR #539, opened 2023-08-24
solution?

sashabryl:


app/test_main.py
```diff
@@ -1,3 +1,53 @@
+import pytest
+
 from app.main import get_human_age
 
-# write your code here
+
+@pytest.mark.parametrize(
+    "dog_age,cat_age,expected_result",
+    [
+        pytest.param(
```
Dimosphen1, line 9: There is no need to use `pytest.param` with `id` here, omit it here, and for error tests.

app/test_main.py
```diff
@@ -1,3 +1,53 @@
...
+            "cat",
+            TypeError,
+            id="should raise TypeError if an argument is not numeric"
+        )
+    ]
+)
+def test_should_raise_type_error_when_wrong_argument(
+        dog_age, cat_age, expected_error) -> None:
+    with pytest.raises(expected_error):
+        get_human_age(dog_age, cat_age)
```
Dimosphen1, line 53: Add test cases for each range of dogs and cats years to cover different function outputs. Besides that, add tests to cover other invalid inputs such as list, dict, set, etc.

Dimosphen1 (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #544, opened 2023-08-25
Solution

CaCtUs1k:


MarianKovalyshyn (commented):
Why you don't test '0', '15' and '24' age. What if at age of 15 you still would get 0 human year or at age of 24 you would get 1 human year. Also it would be good to test negative age and some wrong type of input data (that was in checklist)

MarianKovalyshyn (approved):

app/test_main.py
```diff
@@ -1,3 +1,82 @@
...
+    )
+    def test_get_human_age(
+            self,
+            dog_age: int,
+            cat_age: int,
+            result: list[int]
+    ) -> None:
+        assert get_human_age(dog_age, cat_age) == result
+
+    @pytest.mark.parametrize(
```
IvanRamyk, line 64: I believe it's redundant to create `parametrize` for only one test

IvanRamyk (changes requested):

PR #545, opened 2023-08-25
Solution

OleksandrMoroz24:


app/test_main.py
```diff
@@ -1,3 +1,16 @@
...
+    (0, 0, [0, 0], "zero_age"),
+    (14, 14, [0, 0], "under_15_age"),
+    (15, 15, [1, 1], "15_years"),
+    (23, 23, [1, 1], "under_24_years"),
+    (24, 24, [2, 2], "24_years"),
+    (27, 27, [2, 2], "27_years"),
+    (28, 28, [3, 2], "mixed_years"),
+    (100, 100, [21, 17], "high_age"),
+])
+def test_get_human_age(cat_age, dog_age, expected, test_id):
```
ZaichenkoViktoriia, line 15: you need to add type annotation here, it doesn't work now

ZaichenkoViktoriia (changes requested):

app/test_main.py
```diff
@@ -1,3 +1,57 @@
...
+        (0, 0, [0, 0]),
+        (14, 14, [0, 0]),
+        (15, 15, [1, 1]),
+        (23, 23, [1, 1]),
+        (24, 24, [2, 2]),
+        (27, 27, [2, 2]),
+        (28, 28, [3, 2]),
+        (100, 100, [21, 17])
+    ],
+    ids=[
```
ZaichenkoViktoriia, line 18: it working but I'm not sure that it's a good idea to write IDs in different lists. As for me, this solution will be better when you write IDs in the next string after your arguments for the tests

ZaichenkoViktoriia (changes requested):

app/test_main.py
IvanRamyk: agree with Viktoriia, please fix it :)

IvanRamyk: when ids are put together with args it's more clear what and how you want to test

IvanRamyk (changes requested):

Kroll410 (approved):

PR #546, opened 2023-08-25
Solution py-cat-and-dog-years

katryana:


app/test_main.py
```diff
@@ -1,3 +1,26 @@
...
+@pytest.mark.parametrize(
+    "cat_age,dog_age,expected_result",
+    [
+        (0, 0, [0, 0]),
+        (14, 14, [0, 0]),
+        (15, 15, [1, 1]),
+        (23, 23, [1, 1]),
+        (24, 24, [2, 2]),
+        (27, 27, [2, 2]),
+        (28, 28, [3, 2]),
```
eduardhabryd, line 15: This 6 tests could be replaced with only 3 because they test the same cases.

eduardhabryd (changes requested):
Great Job! However, could be shortened.

app/test_main.py
katryana: These tests check the marginal cases for the boundaries. If they are removed, we cannot be sure whether the years are counted in the same steps as in the task condition, i.e. 15, 9, (4 or 5).

PR #560, opened 2023-08-28
'Solution'

hanovskyiv:


app/test_main.py
```diff
@@ -1,3 +1,51 @@
...
+            result: list
+    ) -> None:
+        assert get_human_age(cat_age, dog_age) == result
+
+    @pytest.mark.parametrize(
+        "cat_age,dog_age,expected_error",
+        [
+            pytest.param(
+                "zero",
+                "zero",
```
Dimosphen1, line 38: You should add tests for other data types as well, not only string (include list, set, dict as well)

app/test_main.py
```diff
@@ -1,3 +1,51 @@
...
-# write your code here
+
+class TestGetHumanAge:
+
+    @pytest.mark.parametrize(
+        "cat_age,dog_age,result",
+        [
+            (0, 0, [0, 0]),
+            (14, 14, [0, 0]),
+            (23, 23, [1, 1]),
```
Dimosphen1, line 13: Add tests when you use different values for `cat_age` and `dog_age` to test that the output is not changed in case different values are received. Besides that, you could cover more ranges with this approach.

app/test_main.py
```diff
@@ -1,3 +1,51 @@
...
+            (100, 100, [21, 17]),
+            (-2, -5, [0, 0])
+        ],
+        ids=[
+            "0 cat/dog years should convert into 0 human age.",
+            "14 cat/dog years should convert into 0 human age.",
+            "23 cat/dog years should convert into 1 human age.",
+            "100 cat/dog years should convert into 21/17 human age.",
+            "negative numbers should convert into 0 human age.",
+        ]
```
Dimosphen1, line 23: There is no need to provide `ids` here.

app/test_main.py
```diff
@@ -1,3 +1,51 @@
...
+            cat_age: int,
+            dog_age: int,
+            result: list
+    ) -> None:
+        assert get_human_age(cat_age, dog_age) == result
+
+    @pytest.mark.parametrize(
+        "cat_age,dog_age,expected_error",
+        [
+            pytest.param(
```
Dimosphen1, line 36: There is no need to use `pytest.param` here.

Dimosphen1 (changes requested):
Several changes are requested.

ArsenPidhoretskyi (approved):
Well done!

PR #562, opened 2023-08-28
solution

ivanzh13:


ivanzh13:
I don't understand what I should do. No function throws an error. How am I supposed to test something that doesn't exist?

app/test_main.py
```diff
@@ -1,3 +1,37 @@
...
+        # test_when_cat_and_dog_years_are_zeros
+        (0, 0, [0, 0]),
+        (-1, -1, [0, 0]),
+        (14, 14, [0, 0]),
+        (15, 15, [1, 1]),
+        (23, 23, [1, 1]),
+        (24, 24, [2, 2]),
+        (27, 27, [2, 2]),
+        (28, 28, [3, 2]),
+        (100, 100, [21, 17])
```
vsmutok, line 18: add a test that will check for invalid data types, for example `"14"`

vsmutok (changes requested):

app/test_main.py
```diff
@@ -1,3 +1,42 @@
...
+    cat_age: int,
+    dog_age: int,
+    expected: list[int]
+) -> None:
+    assert get_human_age(cat_age, dog_age) == expected
+
+
+def test_get_human_age_with_invalid_param_type() -> None:
+    with pytest.raises(TypeError):
+        assert get_human_age("14", "14") == [0, 0]
```
ArsenPidhoretskyi, line 42: Also for example if it accepts [14], [14]
Will be also nice test

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #564, opened 2023-08-29
Solution

WoyDiam:


vasylhnatiuk (changes requested):
https://github.com/mate-academy/py-cat-and-dog-years/blob/master/checklist.md read checklist and add tests for `if the function receives an incorrect type of data it raises the correct exception.`

vasylhnatiuk (changes requested):
create a separate test to test that if the function fails correct exception will be raised. use  with pytest.raises

vasylhnatiuk (changes requested):
do it

vsmutok (approved):
Good Job 👍

PR #578, opened 2023-09-17
Solution

endernaga:


Bohdan-Salamakha (changes requested):
add tests for negative numbers, really big numbers and invalid types

PR #580, opened 2023-09-18
not a solution

d-beschastnyi:
i need your help w/ this error
i just can't understand what i did wrong


vsmutok (changes requested):

Create test for this case

vsmutok (approved):
Good Job 👍

PR #587, opened 2023-09-25
Solution

ladno338:


ArsenPidhoretskyi (changes requested):

vasylhnatiuk (changes requested):

app/test_main.py
```diff
@@ -58,5 +58,5 @@ def test_convert_zero(
 
         assert get_human_age(cat_years, dog_years) == result
 
-    def test_type(self) -> None:
+    def test_type_data(self) -> None:
```
vasylhnatiuk, line 61: i mean that you must add test data for this test and test if this data raise exception

vasylhnatiuk (changes requested):

app/test_main.py
```diff
@@ -1,3 +1,62 @@
...
+            self,
+            cat_years: int,
+            dog_years: int,
+            result: list
+    ) -> None:
+
+        assert get_human_age(cat_years, dog_years) == result
+
+    def test_type_data(self) -> None:
+        assert pytest.raises(TypeError)
```
ArsenPidhoretskyi, line 62: Wrong usage:
https://docs.pytest.org/en/7.1.x/how-to/assert.html#assertions-about-expected-exceptions

ArsenPidhoretskyi (changes requested):

PR #588, opened 2023-09-25
writed tests for main.py

antonnech2309:


viktoria-rybenchuk (changes requested):
What if the function receives incorrect data type or a negative number?

PR #589, opened 2023-09-25
implemented tests to py-cat-and-dog-years task

HlovaOleh:
I wanted to write that I have no idea what I did, but I didn't know if you would get the joke ;)

### DIFF
--- a/app/test_main.py
+++ b/app/test_main.py
@@ -1,3 +1,42 @@
+import pytest
+
 from app.main import get_human_age
 
-# write your code here
+
+class TestGetHumanAge:
+    @pytest.mark.parametrize(
+        "cat_age,dog_age,result",
+        [
+            (0, 0, [0, 0]),
+            (14, 14, [0, 0]),
+            (15, 15, [1, 1]),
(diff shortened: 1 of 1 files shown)


Bohdan-Salamakha (changes requested):
add tests for some negative numbers and really big numbers

app/test_main.py
```diff
@@ -1,3 +1,46 @@
...
+        assert get_human_age(cat_age, dog_age) == result
+
+    @pytest.mark.parametrize(
+        "cat_age,dog_age,expected_error",
+        [
+            ("value", 12, TypeError),
+            (13, "value", TypeError)
+        ]
+    )
+    def test_get_human_age_with_correct_types_of_value(
```
YukihiroSM, line 39: Looks like this test should ne named as something like: "test_get_human_age_with_**in**correct_types_of_value"

YukihiroSM (changes requested):
The code looks good, but there is a question about naming.

YukihiroSM (approved):
Nicely done!

PR #594, opened 2023-10-03
'Solution'

nsdNite:


app/test_main.py
```diff
@@ -1,3 +1,42 @@
+import pytest
```
viktoria-rybenchuk, line 1: fix  ordering imports related to PEP 8  https://peps.python.org/pep-0008/#imports

app/test_main.py
```diff
@@ -1,3 +1,42 @@
...
+        result: list
+) -> None:
+    assert get_human_age(cat_age, dog_age) == result
+
+
+def test_function_input_correct_type() -> None:
+    with pytest.raises(TypeError):
+        get_human_age("42", 42)
+    with pytest.raises(TypeError):
+        get_human_age(42, "42")
```
viktoria-rybenchuk, line 42: DRY. You can use parameterization as well

viktoria-rybenchuk (changes requested):

app/test_main.py
nsdNite: Imports should be grouped in the following order:

Standard library imports.
Related third party imports.
Local application/library specific imports.

Why should i import local application before third party import?

app/test_main.py
```diff
@@ -1,3 +1,51 @@
...
+    [
+        ("42", 42),
+        (42, "42")
+    ],
+    ids=[
+        "cat age is string type",
+        "dog age is string type"
+    ]
+)
+def test_function_input_correct_type(cat_age: int, dog_age: int) -> None:
```
Dimosphen1, line 49: Type annotation for arguments here should be `Any`, there is no guarantee that it would be `int`.

app/test_main.py
```diff
@@ -1,3 +1,51 @@
...
+        result: list
+) -> None:
+    assert get_human_age(cat_age, dog_age) == result
+
+
+@pytest.mark.parametrize(
+    "cat_age, dog_age",
+    [
+        ("42", 42),
+        (42, "42")
```
Dimosphen1, line 42: Add the `TypeError` as the third argument in parametrize that you will pass to a `test_function_input_correct_type`.

app/test_main.py
```diff
@@ -1,3 +1,51 @@
...
+        dog_age: int,
+        result: list
+) -> None:
+    assert get_human_age(cat_age, dog_age) == result
+
+
+@pytest.mark.parametrize(
+    "cat_age, dog_age",
+    [
+        ("42", 42),
```
Dimosphen1, line 41: Add data types to test other than strings, e.g. lists, dicts, sets, tuples.

app/test_main.py
```diff
@@ -1,3 +1,51 @@
...
+    "cat_age, dog_age, result",
+    [
+        (-1, -7, [0, 0]),
+        (0, 0, [0, 0]),
+        (14, 14, [0, 0]),
+        (15, 15, [1, 1]),
+        (23, 23, [1, 1]),
+        (24, 24, [2, 2]),
+        (27, 27, [2, 2]),
+        (28, 28, [3, 2]),
```
Dimosphen1, line 15: Add test cases to cover situations when `cat_age` and `dog_age` have different incoming values as mentioned [in the checklist](https://github.com/nsdNite/py-cat-and-dog-years/blob/develop/checklist.md#clean-code).

Dimosphen1 (changes requested):

pavlejviki (approved):
Well done!

PR #600, opened 2023-10-09
solution

Depaton:


app/test_main.py
```diff
@@ -1,3 +1,17 @@
...
+def test_animal_age_is_less_or_equal_23() -> None:
+    assert get_human_age(23, 23) == [1, 1]
+
+
+def test_animal_is_less_or_equal_27() -> None:
+    assert get_human_age(27, 27) == [2, 2]
+
+
+def test_animal_is_bigger_equal_27() -> None:
+    assert get_human_age(28, 28) == [3, 2]
```
pavlejviki, line 17: use @pytest.mark.parametrize instead of creating multiple tests

pavlejviki (changes requested):

app/test_main.py
```diff
@@ -1,3 +1,34 @@
...
 
-# write your code here
+
+@pytest.mark.parametrize("cat_age, dog_age, output", [
+    (0, 0, [0, 0]),
+    (14, 14, [0, 0]),
+    (15, 15, [1, 1]),
+    (27, 27, [2, 2]),
+    (28, 28, [3, 2]),
+    (28, 20, [3, 1]),
```
Dimosphen1, line 11: Add test cases to test scenarios when values for `cat_age` and `dog_age` differ.

app/test_main.py
```diff
@@ -1,3 +1,34 @@
...
+    (100, 100, [21, 17]),
+    (-10, -10, [0, 0])
+])
+def test_animal_age_with_input_data(
+        cat_age: int,
+        dog_age: int,
+        output: int
```
Dimosphen1, line 18: You do not expect an integer for `output`, but rather a list of integers. Indicate this in your type annotation.

app/test_main.py
```diff
@@ -1,3 +1,34 @@
...
+
+
+@pytest.mark.parametrize("cat_age, dog_age, error", [
+    ("10", "10", TypeError),
+    ([10], [10], TypeError),
+    ({10: ""}, {10: ""}, TypeError),
+])
+def test_animal_age_with_incorrect_data_type(
+        cat_age: str or list or dict,
+        dog_age: str or list or dict,
```
Dimosphen1, line 30: Replace the type annotation for these variables with `Any`, it would be sufficient here.

app/test_main.py
```diff
@@ -1,3 +1,34 @@
 from app.main import get_human_age
+import pytest
```
Dimosphen1, line 2: `pytest` import should precede importing `get_human_age`.

Dimosphen1 (changes requested):

viktoria-rybenchuk (approved):
GJ!

PR #602, opened 2023-10-11
Animal_age_converter

Filllick:


Filllick:
Following the function logit, it will work if it recives data out of normal range, such as negative numbers, zero, or realy large numbers. So I can write the test only to check if recived data is a number.

app/test_main.py
```diff
@@ -1,3 +1,41 @@
...
+        "27 cat and dog ages should return [1, 1]",
+        "28 cat and dog ages should return [3, 2]",
+        "32 cat and 48 dog ages should return [4, 6]",
+        "100 cat and dog ages should return [21, 17]",
+    ]
+)
+def test_animals_to_human_age(
+        cat_age: int,
+        dog_age: int,
+        human_ages: list
```
vasylhnatiuk, line 34:
```suggestion
        human_ages: list[int]
```

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #604, opened 2023-10-12
'Solution'

lylakivan:


app/test_main.py
```diff
@@ -1,3 +1,43 @@
...
+            cat_age: int,
+            dog_age: int,
+            expected_result: list
+    ) -> None:
+        assert get_human_age(cat_age, dog_age) == expected_result
+
+    @pytest.mark.parametrize(
+        "cat_age,dog_age",
+        [
+            (str, int),
```
Dimosphen1, line 31: Provide not data types, but rather values that have these types, e.g. actual strings, lists, floats, e.g.

app/test_main.py
```diff
@@ -1,3 +1,43 @@
...
+            (int, str),
+            (str, str),
+            (None, int),
+            (int, None),
+            (None, None)
+        ]
+    )
+    def test_should_accept_correct_types(
+            self, cat_age: int, dog_age: int
```
Dimosphen1, line 40: Declare here in the type annotation that the incoming value could be `Any` by its type.

app/test_main.py
```diff
@@ -1,3 +1,43 @@
...
+        "cat_age,dog_age,expected_result",
+        [
+            (0, 0, [0, 0]),
+            (14, 14, [0, 0]),
+            (15, 15, [1, 1]),
+            (23, 23, [1, 1]),
+            (24, 24, [2, 2]),
+            (27, 27, [2, 2]),
+            (28, 28, [3, 2]),
+            (100, 100, [21, 17])
```
Dimosphen1, line 17: Add tests to evaluate output when values for `cat_age` and `dog_age` differ from one another or when they are negative.

Dimosphen1 (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #608, opened 2023-10-17
Solution

loplicat:


app/test_main.py
```diff
@@ -1,3 +1,33 @@
...
+    ]
+)
+def test_correct_human_age(
+        cat_age: int,
+        dog_age: int,
+        human_age_of_cat_and_dog: int
+) -> None:
+    assert (get_human_age(cat_age, dog_age) == human_age_of_cat_and_dog), \
+        (f"Human age of {cat_age} cats years and {dog_age} dogs years "
+         f"should be equal to {human_age_of_cat_and_dog}")
```
viktoria-rybenchuk, line 33: Why do you need this? You have already defined test messages above

viktoria-rybenchuk (changes requested):
Read checklist. You should cover all test cases

app/test_main.py
loplicat: Thanks for the comments, I've already fixed the code :)

Dimosphen1: There is no need for the f-string here, remove it and leave only the assert statement that checks that the output equals the `human_age_of_cat_and_dog`. You could rely on the standard error formatting.

app/test_main.py
```diff
@@ -1,3 +1,44 @@
...
+@pytest.mark.parametrize(
+    "cat_age, dog_age, result",
+    [
+        (1, "age", TypeError),
+        (["age"], 1, TypeError),
+        (1, {"dog": 1}, TypeError),
+    ]
+)
+def test_should_raise_error_if_wrong_data_type_provided(
+    cat_age: int, dog_age: int, result: Type[Exception]
```
Dimosphen1, line 41: You could improve the readability if you place each argument on a new line. Besides that, indicate that `cat_age` and `dog_age` could be `Any` in type annotation, their type would not be strict `int`.

Dimosphen1 (changes requested):

pavlejviki (approved):
Well done!

PR #610, opened 2023-10-17
Solution

prochigor:


Naz-iv (approved):
well done

app/test_main.py
```diff
@@ -1,3 +1,79 @@
...
+    [
+        pytest.param(
+            (1, 3),
+            "string",
+            TypeError,
+            id="should return error with not numeric parameters"
+        )
+    ]
+)
+def test_invalid_data_type(
```
fsocie7y, line 73: No need to use parametrize for only one case, u can write single test or add more cases

fsocie7y (changes requested):

LiudmylaKulinchenko (approved):
Good job! 🐞

PR #611, opened 2023-10-18
Done

DimaPr7:


ArsenPidhoretskyi (changes requested):
Instead of duplicate code you can use pytest.parametrize

ArsenPidhoretskyi (approved):
Well done!

PR #612, opened 2023-10-18
Solution

Naz-iv:


mykolamateichuk (approved):

app/test_main.py
```diff
@@ -1,3 +1,102 @@
...
+        [
+            pytest.param(
+                "10",
+                "10",
+                TypeError,
+                id="should raise TypeError if age cat or age dog is not digit"
+            ),
+        ]
+    )
+    def test_expected_errors_raised(
```
fsocie7y, line 80: No need to use parametrize for only one case, u can write single test or add more cases

app/test_main.py
```diff
@@ -1,3 +1,102 @@
...
+                ValueError,
+                id="should raise error whem age < 0"
+            ),
+            pytest.param(
+                999999,
+                999999,
+                ValueError,
+                id="should raise error whem age over expected range"
+            ),
+"""
```
fsocie7y, line 102: u forgot remove this?)

fsocie7y (changes requested):

app/test_main.py
Naz-iv: wasn't sure if we need to test if negative/too high age should through ValueError, so left it there for now before I get advice  during Q&A

LiudmylaKulinchenko (approved):
Good job! 🐞

PR #615, opened 2023-10-18
Solution

vshvanska:


Familenko (changes requested):
What about TypeError https://github.com/mate-academy/py-cat-and-dog-years/blob/master/checklist.md

fsocie7y (approved):
well done ^_^

PR #624, opened 2023-10-19
Solution

katyakoziar:


app/test_main.py
```diff
@@ -1,3 +1,16 @@
+import pytest
 from app.main import get_human_age
 
-# write your code here
+
```
fsocie7y, line 4: Add tests for negative numbers and incorrect types

fsocie7y (changes requested):

app/test_main.py
katyakoziar: Do I need to check the numbers that a function takes or returns?

LiudmylaKulinchenko (approved):
Good job! 🐞

PR #627, opened 2023-10-19
Solution

RabenkoYevhenii:


Lyutillis (approved):

app/test_main.py
```diff
@@ -1,3 +1,84 @@
+import pytest
+
+
 from app.main import get_human_age
 
-# write your code here
+
+def test_should_return_zero_if_age_less_than_15() -> None:
+    assert (get_human_age(14, 14) == [0, 0]
+            ), "Should return zero if age less than 15"
```
LiudmylaKulinchenko, line 9: fix formatting

LiudmylaKulinchenko (changes requested):
No need to use parametrize for only one case, u can write single test or add more cases

LiudmylaKulinchenko (approved):
Good job! 😸

PR #629, opened 2023-10-19
Implement py-cat-and-dog-years tests

SevKrok:


Katherine-Greg (approved):
Good job! 👍

app/test_main.py

LiudmylaKulinchenko (changes requested):

app/test_main.py
```diff
@@ -1,3 +1,33 @@
+import pytest
+
 from app.main import get_human_age
 
-# write your code here
+
```
fsocie7y, line 5: mb better use parametrize?)

fsocie7y (changes requested):

app/test_main.py
SevKrok: No.
here all understandable

fsocie7y (changes requested):

fsocie7y (approved):
u the best 🥇

PR #633, opened 2023-10-19
Solution

Wonsky1:


skoval7 (approved):

app/test_main.py
LiudmylaKulinchenko, line 1: Add one more test case if the function receives an incorrect type of data it raises the correct exception.

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Good job! 😸

PR #639, opened 2023-10-20
Solution

Andrii-Silenko:


Oxbay (changes requested):
You must use ```@pytest.mark.parametrize```
Also u need check func for zero, negative and unreal ages.

Read "checklist" in the end of "Readme"

PR #640, opened 2023-10-20
"Solution"

TROTPOL:


app/test_main.py
```diff
@@ -1,3 +1,26 @@
...
+        "the result is null if the value is null",
+        "values less than 15 don't count",
+        "if the value is from 15 to 24, the result is 1",
+        "if the value is from 25 to 28, the result is 2",
+        "for cat age 28 counts as 3, when for dog it still 2",
+        "for cat age 100 counts as 21, when for dog it still 17"
+    ]
+)
+def test_get_human_age(cat_age: int, dog_age: int, result: list) -> None:
+    assert get_human_age(cat_age, dog_age) == result
```
pavlejviki, line 26: Double check checklist.

if function resive data out of normal range, such as negative numbers;

pavlejviki (changes requested):

app/test_main.py
```diff
@@ -1,3 +1,38 @@
...
+
+
+def test_get_human_age_function_incorrect_type_for_cat() -> None:
+    with pytest.raises(TypeError):
+        get_human_age("1", 1)
+
+
+def test_get_human_age_function_incorrect_type_for_dog() -> None:
+    with pytest.raises(TypeError):
+        get_human_age(1, "1")
```
viktoria-rybenchuk, line 38: Use `parametrize`  to avoid redundancy

viktoria-rybenchuk (changes requested):

pavlejviki (approved):
Well done

PR #641, opened 2023-10-22
Added solution

IlyaGrynyshyn:


app/test_main.py
```diff
@@ -1,3 +1,74 @@
...
+            ),
+            (
+                (28, 4), (28, 1), TypeError
+            )
+        ]
+    )
+    def test_get_human_age_invalid_error(
+            self,
+            cat_age: int,
+            dog_age: int,
```
Dimosphen1, line 70: Change your type annotation to indicate that incoming arguments could be `Any` by their type according to the test setup, it is not guaranteed that they would be integers.

app/test_main.py
```diff
@@ -1,3 +1,74 @@
...
+            error: Exception
```
Dimosphen1, line 71: Indicate that the `error` should be not an instance of `Exception`, but rather a subclass of one.

app/test_main.py
```diff
@@ -1,3 +1,74 @@
...
+            (
+                (),
+                (),
+                TypeError
+            ),
+            (
+                28, (), TypeError
+            ),
+            (
+                (28, 4), (28, 1), TypeError
```
Dimosphen1, line 63: Add tests with datatypes other than tuples for this task (e.g. str, dict, set, list, e.g.).

app/test_main.py
```diff
@@ -1,3 +1,74 @@
...
+                14,
+                14,
+                [0, 0]
+            ),
+            (
+                15, 15, [1, 1]
+            ),
+            (
+                23, 23, [1, 1]
+            ),
```
Dimosphen1, line 20: There is no need for such broad spacing, it affects the code perception. Reformat your code style as follows:
```suggestion
        [
            (14, 14, [0, 0]),
            (15, 15, [1, 1]),
            (23, 23, [1, 1]),
```
Apply this to `test_get_human_age_invalid_error` as well.

app/test_main.py
```diff
@@ -1,3 +1,74 @@
...
+            (
+                0, 0, [0, 0]
+            )
+        ]
+    )
+    def test_get_human_age(
+            self, cat_age: int,
+            dog_age: int,
+            expected_age: list
```
Dimosphen1, line 47: Place each argument on a new line, including `self`. Besides that, indicate that you pass either `int` or `float` for `dog_age` and `cat_age`.

Dimosphen1 (changes requested):

pavlejviki (approved):
Well done

PR #647, opened 2023-10-27
Solution

okien1:


pavlejviki (changes requested):
tests failed

viktoria-rybenchuk (changes requested):

app/main.py
```diff
@@ -1,4 +1,8 @@
 def get_human_age(cat_age: int, dog_age: int) -> list:
+    if cat_age <= 0 or dog_age <= 0:
+        raise ValueError
+    if cat_age >= 100 or dog_age >= 100:
+        raise ValueError
```
Dimosphen1, line 5: You should not modify the main function, work with it in its initial state. Exclude this file from the pull request.

app/test_main.py
```diff
@@ -1,3 +1,80 @@
...
+                28,
+                [3, 2],
+                id="test when ages are equal 28"
+            ),
+            pytest.param(
+                44,
+                44,
+                [7, 6],
+                id="test if function works for third+ year"
+            ),
```
Dimosphen1, line 45: Add test cases to test when the input is a negative number, zero, or a large number.

app/test_main.py
```diff
@@ -1,3 +1,80 @@
...
+                1.6, "6.0", TypeError,
+                id="Throws error if one of ages isn't integer"
+            ),
+            pytest.param(
+                0, 0, ValueError,
+                id="Throws error if ages are equal to zero or less"
+            ),
+            pytest.param(
+                100, 5, ValueError,
+                id="Throws error if one of ages is higher than 99"
```
Dimosphen1, line 69: Exclude these checks from `test_if_function_throws_errors`

app/test_main.py
```diff
@@ -1,3 +1,80 @@
...
+            )
+        ]
+    )
+    def test_if_function_throws_errors(
```
Dimosphen1, line 73: Add more tests for inputs with unexpected types, e.g. lists, tuples, dicts, sets, e.g.

app/test_main.py
```diff
@@ -1,3 +1,80 @@
+import pytest
+
 from app.main import get_human_age
 
-# write your code here
+
+class TestGetHumanAge:
```
Dimosphen1, line 6: There is no need to organize these tests in class, you could write them as two separate functions and omit using `self` as a function argument.

Dimosphen1 (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #648, opened 2023-10-28
Solution

kravetb:


ArsenPidhoretskyi (changes requested):

app/test_main.py
EdAlekseiev, line 1: Make sure you've tests checks all edged situations such as:

if function resive data out of normal range, such as negative numbers;

EdAlekseiev (changes requested):

YukihiroSM (approved):
Good job!

PR #649, opened 2023-10-29
Solution

frezworx:


Polyakiv-Andrey (changes requested):
Tests not passed

app/test_main.py
```diff
@@ -1,3 +1,23 @@
...
+
+@pytest.mark.parametrize(
+    "cat_age, dog_age, expected", [
+        (0, 0, [0, 0]),
+        (14, 14, [0, 0]),
+        (15, 15, [1, 1]),
+        (23, 23, [1, 1]),
+        (24, 24, [2, 2]),
+        (28, 28, [3, 2]),
+        (100, 100, [21, 17]),
```
pavlejviki, line 14: add tests for negative numbers and if the function receives an incorrect type of data it raises the correct exception.

app/test_main.py
```diff
@@ -1,3 +1,23 @@
...
+    ],
+)
+def test_should_convert_into_human_age(
+        cat_age: int,
+        dog_age: int,
+        expected: list
+) -> None:
+    result = get_human_age(cat_age, dog_age)
```
pavlejviki, line 22: no need to create additional variable

pavlejviki (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #650, opened 2023-10-31
solution

Stainback:


YukihiroSM (changes requested):
Add a test to check that function rises the exception in case of wrong input, like "1", "1"

NikOneZ1 (approved):
Well done!

PR #653, opened 2023-11-03
Solution

DanSheremeta:
Solution


EdAlekseiev (changes requested):
Please, add test for the next cases:
1. if function resive data out of normal range, such as negative numbers, zero, or realy large numbers;
2. if the function receives an incorrect type of data it raises the correct exception.

EdAlekseiev (approved):
Good job

PR #655, opened 2023-11-06
Solution

rurakite:


NikOneZ1 (changes requested):
Please, fix tests

EdAlekseiev (approved):
Good job

rurakite:
> Good job

Hello! Sorry, are you sure that all is good there? I mean did I checks all edged situations such as:

Because I had troubles with test the function with a negative number, as the "convert_to_human" function return 0 in case the number is negative. Just want to be clear with that. Thank you!

PR #657, opened 2023-11-08
Solution

hbilous:


EdAlekseiev (changes requested):
Add test for the next cases:

app/test_main.py
```diff
@@ -1,3 +1,29 @@
...
+        cat_age: int,
+        dog_age: int,
+        converted_age: list[int]
+) -> None:
+    assert get_human_age(cat_age, dog_age) == converted_age
+
+
+def test_incorrect_input_type() -> None:
+    with pytest.raises(TypeError):
+        get_human_age("15", "0")
```
YukihiroSM, line 29: Will the function work if only one of them is integer, other is string? Check this as well, please.

app/test_main.py
```diff
@@ -1,3 +1,29 @@
...
+    [
+        (-1, -1, [0, 0]),
+        (0, 0, [0, 0]),
+        (14, 14, [0, 0]),
+        (15, 15, [1, 1]),
+        (23, 23, [1, 1]),
+        (24, 24, [2, 2]),
+        (27, 27, [2, 2]),
+        (28, 28, [3, 2]),
+        (100, 100, [21, 17])
```
YukihiroSM, line 16: Add some huge numbers also, please.

YukihiroSM (changes requested):

EdAlekseiev (approved):
Good job

PR #658, opened 2023-11-09
'Solution'

Illya-Maznitskiy:


app/test_main.py
```diff
@@ -1,3 +1,74 @@
...
+            (23, 25, [1, 2]),
+        ]
+    )
+    def test_additional_cases(
+            self,
+            cat_age: int,
+            dog_age: int,
+            expected: list
+    ) -> None:
+        assert get_human_age(cat_age, dog_age) == expected
```
Dimosphen1, line 74: Group all tests that are aimed to process the integer input to one test, and cases where you test against the raised error to another one. Besides that, add test cases for raising errors in case the input is incorrect, e.g. you pass dicts, lists, sets, bool values as the `cat_age` and `dog_age` arguments.

Dimosphen1 (changes requested):

app/test_main.py
Illya-Maznitskiy: I wrote a tests, but how should I raise the error "TypeError? Help, I can't figure out how to do it

app/test_main.py
```diff
@@ -1,3 +1,110 @@
...
+            (5, 5),
+            (2, 7)
+        ]
+    )
+    def test_integer_inputs(
+            self,
+            cat_age: int,
+            dog_age: int,
+    ) -> None:
+        assert get_human_age(cat_age, dog_age) == [0, 0]
```
pavlejviki, line 92: you can join all the tests above into one sigle test

pavlejviki (changes requested):

app/test_main.py
pavlejviki: And fix tests

vsmutok (approved):
Looks Good 👍

PR #659, opened 2023-11-10
Solution

AnyoneClown:


app/test_main.py
```diff
@@ -1,3 +1,24 @@
...
+    [
+        (0, 0, [0, 0]),
+        (14, 14, [0, 0]),
+        (15, 15, [1, 1]),
+        (23, 23, [1, 1]),
+        (24, 24, [2, 2]),
+        (27, 27, [2, 2]),
+        (28, 28, [3, 2]),
+        (100, 100, [21, 17]),
+        (-1, -1, [0, 0])
```
pavlejviki, line 16: add tests to check if the function receives an incorrect type of data it raises the correct exception.

pavlejviki (changes requested):

vsmutok (approved):
looks Good 👍

PR #660, opened 2023-11-12
Implemented py-cat-and-dog-years

Aeros72:


viktoria-rybenchuk (changes requested):
How do you handle if the function receives an incorrect type of data it raises the correct exception? Read checklist

app/test_main.py
```diff
@@ -1,3 +1,31 @@
...
+        cat_age: int,
+        dog_age: int,
+        expected: list
+) -> None:
+    assert get_human_age(cat_age, dog_age) == expected
+
+
+def test_incorrect_input_type() -> None:
+    with pytest.raises(TypeError):
+        assert get_human_age("16", 15)
```
pavlejviki, line 31: better to add more test cases using @pytest.mark.parametrize

pavlejviki (changes requested):

viktoria-rybenchuk (approved):
Good job!

PR #662, opened 2023-11-15
Solution

Taran21V:


NikOneZ1 (changes requested):
Please include tests for the following scenarios:

1. If the function receives data outside the normal range, such as negative numbers, zero, or extremely large numbers.
2. If the function receives data of an incorrect type, it should raise the appropriate exception.

PR #671, opened 2023-11-20
Solution

Artem-user-19:


app/test_main.py
```diff
@@ -1,3 +1,16 @@
+import pytest
 from app.main import get_human_age
```
Dimosphen1, line 2: Make one blank line between the different groups of imports according to PEP-8.

app/test_main.py
```diff
@@ -1,3 +1,16 @@
...
+
+@pytest.mark.parametrize("cat_age, dog_age, expected", [
+    (0, 0, [0, 0]),
+    (14, 14, [0, 0]),
+    (15, 15, [1, 1]),
+    (23, 23, [1, 1]),
+    (24, 24, [2, 2]),
+    (27, 27, [2, 2]),
+    (28, 28, [3, 2]),
+    (100, 100, [21, 17]),
```
Dimosphen1, line 13: Add a test case for negative values and a test case with different values for `cat_age` and `dog_age` within one test to cover the requirements of the [checklist](https://github.com/Artem-user-19/py-cat-and-dog-years/blob/develop/checklist.md#clean-code).

app/test_main.py
```diff
@@ -1,3 +1,16 @@
...
+])
+def test_get_human_age(cat_age: int, dog_age: int, expected: int) -> None:
+    assert get_human_age(cat_age, dog_age) == expected
```
Dimosphen1, line 16: Add a separate test to test `get_human_age` in case it receives unexpected arguments with data types other than integers. You should make sure that it raises the correct exception according to the [checklist](https://github.com/Artem-user-19/py-cat-and-dog-years/blob/develop/checklist.md#clean-code).

Dimosphen1 (changes requested):

app/test_main.py
```diff
@@ -1,3 +1,31 @@
...
+        expected: int
+) -> None:
+    assert get_human_age(cat_age, dog_age) == expected
+
+
+def test_get_human_age_invalid() -> None:
+    # Use pytest.raises to check that the function raises the correct exception
+    with pytest.raises(TypeError):
+        # Pass a string as an argument
+        get_human_age("cat", "dog")
```
Dimosphen1, line 31: Use `pytest.mark.parametrize` for this test as well. Test against different data types other than strings, e.g. against tuples, dicts, lists, etc.

app/test_main.py
```diff
@@ -1,3 +1,31 @@
...
+    (27, 27, [2, 2]),
+    (28, 28, [3, 2]),
+    (100, 100, [21, 17]),
+    (-1, -1, [0, 0]),
+    (15, 24, [1, 2]),
+])
+def test_get_human_age_positive(
+        cat_age: int,
+        dog_age: int,
+        expected: int
```
Dimosphen1, line 22: Change your type annotation for `expected`, it is not an `int`.

app/test_main.py
Dimosphen1, line 31: Remove comments from your code, there is no need for them.

Dimosphen1 (changes requested):

viktoria-rybenchuk (approved):
Goof job!

PR #672, opened 2023-11-22
Solution

DenVia:


app/test_main.py

LiudmylaKulinchenko (changes requested):

app/test_main.py
```diff
@@ -1,3 +1,25 @@
...
+    (28, 29, [3, 3]),
+    (100, 100, [21, 17])
+])
+def test_get_human_age(
+        cat_age: int,
+        dog_age: int,
+        expected_result: list
+) -> None:
+    if not isinstance(cat_age, int) or not isinstance(dog_age, int):
+        raise TypeError("Arguments should be int type")
```
LiudmylaKulinchenko, line 23: You should create a new test for incorrect types fot the tested function, not make actions in your test

LiudmylaKulinchenko (changes requested):

PR #673, opened 2023-11-22
Solution

lyaskov:


YukihiroSM (changes requested):
Your tests are good, but can you also add tests, that function raises error, when the wrong value is passed (string for example), also, test with really big numbers.

EdAlekseiev (approved):
Good job

PR #676, opened 2023-11-23
Solution

mrve4nick:


app/test_main.py
```diff
@@ -1,3 +1,50 @@
...
+            ),
+        ]
+    )
+    def test_get_human_age(
+            self,
+            cat_age: int,
+            dog_age: int,
+            expected: list[int]
+    ) -> None:
+        assert get_human_age(cat_age, dog_age) == expected
```
Dimosphen1, line 50: Not all test cases from the checklist are covered. You should add cases to test `get_human_age` against negative numbers, different numbers for `cat_age` and `dog_age` to make sure that output does not change when this difference is present, and tests with data types for `cat_age` and `dog_age` different from `int` (e.g. tuples, list, string, etc.). The last test cases should test that the specific error is raised during the function execution.

Dimosphen1 (changes requested):
Additional test cases were requested.

app/test_main.py
```diff
@@ -1,3 +1,63 @@
...
+
+    def test_get_human_age_with_invalid_type(self) -> None:
```
pavlejviki, line 61: add more test cases for this test using a parametrizer

pavlejviki (changes requested):

pavlejviki (approved):
Well done!

PR #680, opened 2023-11-28
Cat_ans_dogs_age

MykolaMazh:


Polyakiv-Andrey (changes requested):

pavlejviki (changes requested):
Double check checklist. Add test cases

if function receives data out of normal range, such as negative numbers, zero, or realy large numbers;

MykolaMazh:
> Double check checklist. Add test cases
> 
> if function receives data out of normal range, such as negative numbers, zero, or realy large numbers; if the function receives an incorrect type of data it raises the correct exception.

As I see it when function receives negative or enormous numbers it should raise an exception and then
```python
    @pytest.mark.parametrize(
        "cat_age,dog_age,expected_error",
        [
            pytest.param(
                -10, -42, ValueError,
                id="raise an error with negative initial data"
            ),
            pytest.param(
                120, 120, ValueError,
                id="raise an error with unlikely initial data"
            ),
        ]
    )
    def test_raising_errors(
            self,
            cat_age: int,
            dog_age: int,
            expected_error: Exception
    ) -> None:
```
but with that
image

so i assume i have to  write tests  the function must pass

And here
image
is being told that one should not test case of incorrect data types being passed cause they are specified in the function itself

viktoria-rybenchuk (approved):
Good job!

PR #681, opened 2023-11-29
Solution for "py_cat_and_dog_years"

wataschi:


pavlejviki (changes requested):
Double check ckecklist and add testcases

if function resive data out of normal range, such as negative numbers

app/test_main.py
```diff
@@ -1,3 +1,95 @@
...
+            self,
+            initial_cat_age: Any,
+            initial_dog_age: Any,
+            expected_error: Any
+    ) -> None:
+        with pytest.raises(expected_error):
+            get_human_age(initial_cat_age, initial_dog_age)
+
+    def test_get_error_if_negative_ages(self) -> None:
+        assert get_human_age(-100, -100) == [0, 0]
```
viktoria-rybenchuk, line 95: no need create separate function for this test. You can add this data to `@pytest.mark.parametrize` for testing

app/test_main.py
```diff
@@ -1,3 +1,95 @@
+import pytest
+from typing import Any
 from app.main import get_human_age
```
viktoria-rybenchuk, line 3: fix imports. You should put a blank line between each group of the imports.  You can find info here https://peps.python.org/pep-0008/#imports
Screenshot from 2023-12-06 19-07-47

viktoria-rybenchuk (changes requested):

Dimosphen1 (approved):
LGTM, one comment is left for consideration.

PR #683, opened 2023-11-30
Solution

maxvell4:


viktoria-rybenchuk (changes requested):

Use `@pytest.mark.parametrize`


Read checklist before pushing your solution

viktoria-rybenchuk (approved):
Good job!

PR #690, opened 2023-12-06
Solution

Liliyakhu:


app/test_main.py
```diff
@@ -1,3 +1,88 @@
...
 from app.main import get_human_age
 
-# write your code here
+
+class TestGetHumanAge:
+    @pytest.mark.parametrize(
+        "initial_cat_age,"
+        "initial_dog_age,"
+        "expected_human_age_cat,"
+        "expected_human_age_dog",
```
Dimosphen1, line 10: Unite the `expected_human_age_cat` and `expected_human_age_dog` to one variable `expected_result` where you will store a list of values. It will help to shorten the space and make the code more concise.

app/test_main.py
```diff
@@ -1,3 +1,88 @@
...
+        [
+            pytest.param(
+                0,
+                0,
+                0,
+                0,
+                id="test should return zero when age is zero"
```
Dimosphen1, line 17: Rewrite input values as follows:
```suggestion
                0,
                0,
                [0, 0],
                id="test should return zero when age is zero"
```

app/test_main.py
```diff
@@ -1,3 +1,88 @@
...
+            pytest.param(
+                -4,
+                -3,
+                0,
+                0,
+                id="test should return zero when "
+                   "age is less then zero"
+            )
+        ]
+    )
```
Dimosphen1, line 59: Add test cases with really large values and positive values for `initial_cat_age` and `initial_dog_age` that differ from one another according to the checklist.

app/test_main.py
```diff
@@ -1,3 +1,88 @@
...
+                "5",
+                TypeError,
+                id="should raise error when age is not 'int'"
+            )
+        ]
+    )
+    def test_raising_errors_correctly(
+            self,
+            initial_cat_age: int,
+            initial_dog_age: int,
```
Dimosphen1, line 84: Change your type annotation to indicate that `initial_cat_age` and `initial_dog_age` could be `Any` in this test, not explicitly `int`.

app/test_main.py
```diff
@@ -1,3 +1,88 @@
...
+            expected_error: TypeError
+    ) -> None:
+        with pytest.raises(Exception):
```
Dimosphen1, line 87: Pass here `expected_error`, not `Exception`.

app/test_main.py
```diff
@@ -1,3 +1,88 @@
...
+                TypeError,
+                id="should raise error when age is not 'int'"
+            )
+        ]
+    )
+    def test_raising_errors_correctly(
+            self,
+            initial_cat_age: int,
+            initial_dog_age: int,
+            expected_error: TypeError
```
Dimosphen1, line 85: Add more test cases to test against different data types as input values, e.g. dicts, lists, sets, tuples, etc., not only strings.

app/test_main.py
```diff
@@ -1,3 +1,88 @@
+import pytest
 from app.main import get_human_age
```
Dimosphen1, line 2: Add a blank line between different groups of imports according to PEP-8.

Dimosphen1 (changes requested):

PR #696, opened 2023-12-18
implemented tests for get_human-age function

380974261314kv:


app/test_main.py
```diff
@@ -1,3 +1,32 @@
...
+def test_get_human_age(
+        cat_age: int,
+        dog_age: int,
+        result: list
+) -> None:
+
+    assert (
+        get_human_age(cat_age, dog_age) == result
+    ), (f"When cats_age equal {cat_age}, "
+        f"dogs_age equal {dog_age} result should be {result}")
```
pavlejviki, line 32: as well you need to test  if the function receives an incorrect type of data it raises the correct exception.

pavlejviki (changes requested):

app/test_main.py
```diff
@@ -1,3 +1,56 @@
...
+        assert (
+            get_human_age(cat_age, dog_age) == result
+        ), (f"When cats_age equal {cat_age}, "
+            f"dogs_age equal {dog_age} result should be {result}")
+
+    @pytest.mark.parametrize(
+        "cat_age, dog_age, expected_error",
+        [
+            pytest.param("2", 8, TypeError,
+                         id="expected type value - integer not str"),
```
pavlejviki, line 46: add more test cases

app/test_main.py
```diff
@@ -1,3 +1,56 @@
...
+        ]
+    )
+    def test_get_human_age_with_incorrect_values(
+            self,
+            cat_age: int,
+            dog_age: int,
+            expected_error: type
```
pavlejviki, line 53: wrong type annotation for expected_error

pavlejviki (changes requested):

app/test_main.py
```diff
@@ -1,3 +1,60 @@
...
+            pytest.param(9, [1], TypeError,
+                         id="value should be integer"),
+            pytest.param("1", "1", TypeError,
+                         id="both values must be integer")
+        ]
+    )
+    def test_get_human_age_with_incorrect_values(
+            self,
+            cat_age: int,
+            dog_age: int,
```
Dimosphen1, line 56: You should not indicate that `cat_age` and `dog_age` are `int`, they could be `Any` in the type annotation according to the test setup. Pay attention to this in the future.

Dimosphen1 (approved):
LGTM, one comment is left for consideration.

PR #697, opened 2023-12-18
added tests for convert_to_human func

BGBroth:


app/test_main.py
```diff
@@ -1,3 +1,33 @@
...
+    (15, 15, [1, 1]),
+    (23, 23, [1, 1]),
+    (24, 24, [2, 2]),
+    (27, 27, [2, 2]),
+    (28, 28, [3, 2]),
+    (100, 100, [21, 17]),
+    (-1, 10, [0, 0]),
+    (10, -1, [0, 0]),
+    (0, 10, [0, 0]),
+    (10, 0, [0, 0]),
```
Dimosphen1, line 18: Add test cases to cover scenartios when `cat_age` and `dog_age` values differ to check that the output does not depend on this difference.

app/test_main.py
```diff
@@ -1,3 +1,33 @@
...
+    (1000, 1000, [246, 197]),
+    ("cat", 10, TypeError),
+    (10, "dog", TypeError),
```
Dimosphen1, line 21: Add test cases for errors for inputs other than strings, e.g. lists, sets, tuples, etc.

app/test_main.py
```diff
@@ -1,3 +1,33 @@
...
+        cat_age: int,
+        dog_age: int,
+        expected_result: Any
+) -> None:
+    if expected_result == TypeError:
+        with pytest.raises(TypeError):
+            get_human_age(cat_age, dog_age)
+    else:
+        result = get_human_age(cat_age, dog_age)
+        assert result == expected_result
```
Dimosphen1, line 33: You could split the logic of tests between the two different functions to test against expected values and the error-prone inputs.

app/test_main.py
```diff
@@ -1,3 +1,33 @@
+import pytest
 from app.main import get_human_age
+from typing import Any
```
Dimosphen1, line 3: Order your imports and separate different groups of them with one blank line according to PEP-8.

Dimosphen1 (changes requested):

PR #699, opened 2023-12-22
ready solution!

job-artem:


app/test_main.py
```diff
@@ -1,3 +1,61 @@
...
+    [
+        (100, 100, [21, 17])
+    ]
+)
+def test_every_4_last_years_for_dog(
+        cat_age: int, dog_age: int, result: list
+) -> None:
+    assert (
+        get_human_age(cat_age, dog_age) == result
+    ), "the next 4 cat years after first 24 give 1 more human year"
```
Dimosphen1, line 61: Do not break up these tests to four different functions, you could include them to `pytest.mark.parametrize` at once since the logic is not changed when you pass integers. Besides that, add test cases to check a function with negative input and input with different values for `cat_age` and `dog_age` according to the checklist. You could create a separate function to test against raised errors in case your input is different from integers (e.g. strings, dicts, lists, sets, etc.).

app/test_main.py
```diff
@@ -1,3 +1,61 @@
...
+
+
+@pytest.mark.parametrize(
+    "cat_age,dog_age,result",
+    [
+        (28, 28, [3, 2])
+    ]
+)
+def test_every_4_last_years_for_cat(
+        cat_age: int, dog_age: int, result: list
```
Dimosphen1, line 43: It would be better to place each argument on a new line to improve the readability.

Dimosphen1 (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #700, opened 2023-12-25
all tests passed

bkirichenko:


vasylhnatiuk (changes requested):
Use @pytest.mark.parametrize

vsmutok (changes requested):
Screenshot 2023-12-25 at 20 41 04

app/test_main.py
```diff
@@ -1,3 +1,75 @@
...
+            pytest.param(
+                10000,
+                7000,
+                [2496, 1397],
+                id="should work correctly with very large input values"
+            )
+
+        ]
+    )
+    def test_age(self, dog_age: int, cat_age: int, humane_age: list) -> None:
```
vasylhnatiuk, line 48:
```suggestion
    def test_age(self, dog_age: int, cat_age: int, humane_age: list[int]) -> None:
```

vasylhnatiuk (changes requested):

PR #709, opened 2024-01-06
'Solution'

yablonskyiv:


YukihiroSM (changes requested):
Please, use @pytest.mark.parametrize instead of creating several same tests. 
Also, please, add a test, that function raises exception in case if arguments are in wrong type, for example, if we use it like: get_human_age("10", 10)

app/test_main.py
```diff
@@ -1,3 +1,28 @@
...
+    (23, 23, [1, 1]),
+    (24, 24, [2, 2]),
+    (27, 27, [2, 2]),
+    (28, 28, [3, 2]),
+    (100, 100, [21, 17]),
+])
+def test_get_human_age(
+        cat_age: int,
+        dog_age: int,
+        expected_result: list
```
vasylhnatiuk, line 18:
```suggestion
        expected_result: list[int]
```

vasylhnatiuk (changes requested):

PR #718, opened 2024-01-14
py-cat-and-dog-years

Vladyslav-Pliashechnyk:


msymonovych:
You should check will function raise correct exception if receives an incorect type of data.
You can see it in checkllist

msymonovych (approved):

Vladyslav-Pliashechnyk:
> You should check will function raise correct exception if receives an incorect type of data. You can see it in checkllist

I guess, the test which check if function raises correct exception is excessive, because annotations are indicated.

Vladyslav-Pliashechnyk:
> You should check will function raise correct exception if receives an incorect type of data. You can see it in checkllist

I guess, the test which check if function raises correct exception is excessive, because annotations are indicated.

NEkropo1:
> > You should check will function raise correct exception if receives an incorect type of data. You can see it in checkllist
> 
> I guess, the test which check if function raises correct exception is excessive, because annotations are indicated.

I agree with you, but sometimes we have business logic, which should be checked. I know it's excessive for this current case, but idea is, that dev could implement soething, which breaks default logic, and we should be ensured, that it's not broken, that's why we have that mention in checklist, always follow the checklist )

app/test_main.py
```diff
@@ -1,3 +1,24 @@
...
+        (100, 100, [21, 17]),
+        (-1, -1, [0, 0]),
+    ]
+)
+def test_should_return_correct_list(
+    cat_age: int,
+    dog_age: int,
+    expected_list: list
+) -> None:
+    assert get_human_age(cat_age, dog_age) == expected_list
```
NEkropo1, line 24: always follow the checklist! Also you could add some explicit messages to edge cases, to make them more understandable:
```suggestion
import pytest

@pytest.mark.parametrize(
    "cat_age, dog_age, expected_list",
    [
        (0, 0, [0, 0]),
        (14, 14, [0, 0]),
        (15, 15, [1, 1]),
        (23, 23, [1, 1]),
        (24, 24, [2, 2]),
        (27, 27, [2, 2]),
        (28, 28, [3, 2]),
        (100, 100, [21, 17]),
        (-1, -1, [0, 0]),
    ],
    ids=[
        "edge case: zero ages",
        "edge case: boundary before conversion changes",
        "boundary age for conversion to 1 human year",
        "boundary age just before next conversion",
        "age at second conversion to human years",
        "age just before uneven conversion for cat and dog",
        "age at uneven conversion for cat and dog",
        "large age numbers",
        "negative ages"
    ]
)
def test_should_return_correct_list(
    cat_age: int,
    dog_age: int,
    expected_list: list
) -> None:
    assert get_human_age(cat_age, dog_age) == expected_list

```

NEkropo1 (changes requested):

PR #721, opened 2024-01-15
Solution

kabust:


NEkropo1 (changes requested):
try it with parametrize and messages, explaining the case

NEkropo1 (approved):
Good job!

PR #726, opened 2024-01-18
Solution

pnasonov:


vasylhnatiuk (changes requested):
f the function receives an incorrect type of data it raises the correct exception.

    Use @pytest.mark.parametrize

PR #731, opened 2024-01-21
Solution

u123dev:


vasylhnatiuk (changes requested):
add test for if the function receives an incorrect type of data it raises the correct exception.

app/test_main.py
```diff
@@ -1,3 +1,44 @@
...
+        get_human_age(3, "3")
+
+    with pytest.raises(TypeError):
+        get_human_age([3], 3)
+
+    with pytest.raises(TypeError):
+        get_human_age(3, {3})
+
+    with pytest.raises(TypeError):
+        get_human_age((3, 3))
```
EdAlekseiev, line 44: Use `pytest.mark.parametrize`

EdAlekseiev (changes requested):

vasylhnatiuk (approved):

PR #732, opened 2024-01-21
Solution

MaxymChyncha:


app/test_main.py
```diff
@@ -1,3 +1,29 @@
...
+    def test_human_age_correctly(
+            self,
+            cat_age: int,
+            dog_age: int,
+            expected_result: list[int]
+    ) -> None:
+        assert get_human_age(cat_age, dog_age) == expected_result, (
+            f"For cat age {cat_age} and dog age {dog_age} "
+            f"result should be {expected_result}"
+        )
```
NEkropo1, line 29: Since you cover corner cases, I suggest to try more detalized messages as 1 more parameter

NEkropo1 (changes requested):

app/test_main.py
```diff
@@ -16,6 +16,18 @@ class TestGetHumanAge:
...
+            "Should be zeros if years less than 15",
+            "Should be ones if years equal to 15",
+            "Should be ones if years less than 24",
+            "Should be twos if years equal to 24",
+            "Should be twos if years less than 28",
+            "Should be three for cat years and two for dog years "
+            "if years equal to 28",
+            "Should be 21 for cat years and 17 dog years "
+            "if years equal to 100",
+        ]
```
NEkropo1, line 30: I meant not just mentioning it, but explaining why, like: "Newborn cats and dogs should be equivalent to 0 human years"
Well, still good job!

NEkropo1 (approved):
Well done!

PR #733, opened 2024-01-22
Cat and Dog years

Ananiev-Vitalii:


vasylhnatiuk (changes requested):
write test for 
`if the function receives an incorrect type of data it raises the correct exception.`

PR #737, opened 2024-01-23
Solution

Meksim-fesh:


app/test_main.py
```diff
@@ -1,3 +1,54 @@
...
+        ),
+        pytest.param(
+            80, 80, [16, 13],
+            id="should return [16, 13] when ages are greater than \
+                first_year + second_year + each_year"
+        )
+    ]
+)
+def test_should_return_correct_result(
+    cat_age: int, dog_age: int, expected_result: int
```
vasylhnatiuk, line 52:
```suggestion
    cat_age: int, dog_age: int, expected_result: list[int]
```

vasylhnatiuk (changes requested):
do test for - if the function receives an incorrect type of data it raises the correct exception.

vasylhnatiuk (approved):

PR #741, opened 2024-01-24
Solution

barabarinov:


PythonZem (approved):
Gj!

app/test_main.py
```diff
@@ -34,3 +36,25 @@ def test_get_human_age(
...
+    ids=[
+        "Cat age should be Integer, not String",
+        "Dog age should be Integer, not String",
+        "Cat age and Dog age should be Integers, not Strings",
+    ],
+)
+def test_get_human_age_invalid_input(
+    cat_age: int,
+    dog_age: int,
+    expected_error: Exception,
```
NEkropo1, line 57:
```suggestion
def test_get_human_age_invalid_input(
    cat_age: int,
    dog_age: int,
    expected_error: Type[Exception],
```

app/test_main.py
```diff
@@ -34,3 +36,25 @@ def test_get_human_age(
...
+        (24, "24", TypeError),
+        ("28", "29", TypeError),
+    ],
+    ids=[
+        "Cat age should be Integer, not String",
+        "Dog age should be Integer, not String",
+        "Cat age and Dog age should be Integers, not Strings",
+    ],
+)
+def test_get_human_age_invalid_input(
```
NEkropo1, line 54:
```suggestion
def test_get_human_age_invalid_input_type
```
It's ok name, but as for now it checks only type, so better to explicit mention it.

NEkropo1 (changes requested):

PR #742, opened 2024-01-25
Solved

Dmitriy-Poplinski:


pavlejviki (changes requested):
tests failed

app/test_main.py
```diff
@@ -1,3 +1,28 @@
...
+
+
+@pytest.mark.parametrize(
+    "cat_age,dog_age,expected_human_age",
+    [
+        (0, 0, [0, 0]),
+        (16, 16, [1, 1]),
+        (25, 28, [2, 2]),
+        (14, 14, [0, 0]),
+        (44, 48, [7, 6]),
```
pavlejviki, line 13: You should cover all test cases

if function receives data out of normal range, such as negative numbers  or really large numbers;

pavlejviki (changes requested):

app/main.py
```diff
@@ -1,4 +1,10 @@
 def get_human_age(cat_age: int, dog_age: int) -> list:
```
vasylhnatiuk, line 1: Don’t change main function! You should write only tests

vasylhnatiuk (changes requested):

vsmutok (approved):
Great!

PR #751, opened 2024-01-28
Solution

mdubyna:


Dobryk12 (approved):

app/test_main.py
```diff
@@ -1,3 +1,41 @@
+import pytest
+
 from app.main import get_human_age
 
-# write your code here
+
+@pytest.mark.parametrize(
+    "cat_age, dog_age, result",
+    [
+        (14, 14, [0, 0]),
```
NEkropo1, line 9: I'd prefer to pass ids directly into parametrize args, like :
```suggestion
    "cat_years, dog_years, expected, test_id",
    [
        (0, 0, [0, 0], "0 cat/dog should convert into [0, 0]"),
    ]
)
```
so we can just throw it as a message in assertion after?

app/test_main.py
```diff
@@ -1,3 +1,41 @@
...
+        (100, 100, [21, 17]),
+        (-5, -10, [0, 0]),
+    ],
+    ids=[
+        "age ge 0 and le 14 should equal 0",
+        "age ge 15 and le 23 should equal 1",
+        "age ge 24 and le 27 should equal 2",
+        "age ge 28 should depend on the type of animal",
+        "age ge 28 should depend on the type of animal",
+        "negative age is impossible, should return 0"
```
NEkropo1, line 22: ge and le isn't that much descriptive afm )

NEkropo1 (changes requested):

app/test_main.py
mdubyna: yes we can and I like this option, but why do we need the id parameter?

PR #753, opened 2024-01-28
Solution

achernyshovv:


mdubyna (changes requested):
u need to use  @pytest.mark.parametrize

app/test_main.py
```diff
@@ -1,3 +1,21 @@
...
+
+@pytest.mark.parametrize("cat_age, dog_age, expected_result", [
+    (0, 0, [0, 0]),
+    (14, 14, [0, 0]),
+    (15, 15, [1, 1]),
+    (24, 24, [2, 2]),
+    (28, 28, [3, 2]),
+    (100, 100, [21, 17]),
+    (23, 23, [1, 1]),
+    (27, 27, [2, 2])
```
NEkropo1, line 14: some explanation of why we should get these results?

NEkropo1 (changes requested):
Also always follow checklist, you didn't finish the task, if it's not following the checklist.

PR #754, opened 2024-01-29
Solution

oksankakot:


VladKyianytsia (changes requested):
You should check that it raises the correct exception if the function receives an incorrect type of data.

VladKyianytsia (approved):

NEkropo1 (approved):

NEkropo1 (approved):
Good job!

PR #756, opened 2024-01-29
Solution

kstorozhenko:


MaxOperchuk (approved):
You need to test if the function receives an incorrect type of data.
You may read it in checklist.

PR #757, opened 2024-01-29
'Solution'

MaxOperchuk:


NataliiaKorzh (approved):

app/test_main.py
```diff
@@ -1,3 +1,65 @@
...
+            27, 28, [2, 2]),
+        pytest.param(
+            28, 29, [3, 3]),
+        pytest.param(
+            100, 100, [21, 17])
+    ],
+    ids=[
+        "should return [0, 0] if cat and dog age < 15",
+        "should return [0, 0] if cat and dog age < 15",
+        "should return [0, 0] if cat and dog age < 15",
```
NEkropo1, line 30:
```suggestion
@pytest.mark.parametrize(
    "cat_years, dog_years, expected, test_id",
    [
        (0, 0, [0, 0], "0 cat/dog should convert into [0, 0]"),
    ]
)
def test_get_human_age(cat_years, dog_years, expected, test_id):
    result = get_human_age(cat_years, dog_years)
    assert result == expected, test_id
```
Try this code style.
Also check corner cases with [2, 3], which is not covered.

NEkropo1 (changes requested):

PR #758, opened 2024-01-29
'Solution'

NataliiaKorzh:


app/test_main.py
```diff
@@ -1,3 +1,21 @@
...
+        (0, 0, [0, 0]),
+        (14, 14, [0, 0]),
+        (15, 15, [1, 1]),
+        (23, 23, [1, 1]),
+        (24, 24, [2, 2]),
+        (27, 27, [2, 2]),
+        (28, 28, [3, 2]),
+        (95, 95, [19, 16]),
+    ]
+)
```
shaiduchyk, line 19: image

shaiduchyk (changes requested):

app/test_main.py
```diff
@@ -6,15 +6,42 @@
...
+            id="should return [0, 0] if cat and dog age is 0"
+        ),
+        pytest.param(
+            14, 14, [0, 0],
+            id="should return [0, 0] if cat and dog age is 14"
+        ),
+        pytest.param(
+            15, 15, [1, 1],
+            id="should return [1, 1] if cat and dog age is 15"
+        ),
```
NEkropo1, line 24: You can throw them all in parametrize, like(just a dummy):
```python
@pytest.mark.parametrize(
    [
    ]
)
def test_get_human_age(cat_years, dog_years, expected, test_id):
    result = get_human_age(cat_years, dog_years)
    assert result == expected, test_id
```

NEkropo1 (changes requested):
Also check if your code is according to checklist, since now you didn't fulfill the task

PR #759, opened 2024-01-29
Task solution

shaiduchyk:


aLEKS-e3 (approved):
I'd recommend you check negative values since the `get_human_age` function handles that range. But it's GOOD overall!

NEkropo1 (changes requested):
Always look if your code is according to checklist.

NEkropo1 (changes requested):
negative values not checked, still checklist.

PR #775, opened 2024-02-05
'Solution'

sofiazholud:


app/test_main.py
```diff
@@ -1,3 +1,12 @@
...
+
+def test_get_human_age() -> None:
+    assert get_human_age(0, 0) == [0, 0]
+    assert get_human_age(14, 14) == [0, 0]
+    assert get_human_age(15, 15) == [1, 1]
+    assert get_human_age(23, 23) == [1, 1]
+    assert get_human_age(24, 24) == [2, 2]
+    assert get_human_age(27, 27) == [2, 2]
+    assert get_human_age(28, 28) == [3, 2]
+    assert get_human_age(100, 100) == [21, 17]
```
EdAlekseiev, line 12: Use pytest.mark.parametrize

EdAlekseiev (changes requested):
Add tests for the next cases:

app/test_main.py
sofiazholud: done

app/test_main.py
```diff
@@ -1,3 +1,36 @@
...
+        (27, 27, [2, 2]),
+        (28, 28, [3, 2]),
+        (100, 100, [21, 17]),
+        (-1, 0, [0, 0]),
+        (0, -1, [0, 0]),
+        (-1, -1, [0, 0]),
+        (1000, 1000, [246, 197])
+    ],
+)
+def test_get_human_age(cat_age: int, dog_age: int, expected: list) -> None:
```
vasylhnatiuk, line 22:
```suggestion
def test_get_human_age(cat_age: int, dog_age: int, expected: list[int]) -> None:
```

vasylhnatiuk (changes requested):

app/test_main.py
sofiazholud: done

vasylhnatiuk (approved):

PR #778, opened 2024-02-06
Solution

AlexGrytsai:


vasylhnatiuk (changes requested):
write test to check if the function receives an incorrect type of data it raises the correct exception.

EdAlekseiev (approved):
Good job

PR #782, opened 2024-02-07
Solution

cth-usq:


app/test_main.py
```diff
@@ -1,3 +1,52 @@
...
+    )
+    def test_get_human_age_error(
+            self,
+            cat_age: int,
+            dog_age: int,
+            expected_error: TypeError
+    ) -> None:
+
+        with pytest.raises(TypeError):
+            get_human_age(cat_age, dog_age)
```
NazarSheremet, line 52: this test doesn't use variables expected_error, change it
```suggestion
        with pytest.raises(expected_error):
            get_human_age(cat_age, dog_age)
```

NazarSheremet (changes requested):
Small changes

app/test_main.py
cth-usq: Got it. 
Done!

NazarSheremet (approved):
Good job, changes are correct

PR #785, opened 2024-02-09
Solution

volry:


NazarSheremet (changes requested):
In this this task better to use  `@pytest.mark.parametrize`

volry:
yes, it's best approach. Thought about it )

NazarSheremet (approved):
Good job!

PR #793, opened 2024-02-15
Solution

OleksandrMV:


YukihiroSM (changes requested):
Could you please add more tests, where you would check, if the function raises exceptions in case of wrong parameters were passed (for example, string instead of int)

PR #802, opened 2024-02-22
Solution

VasylRybalka:


EdAlekseiev (changes requested):
Add tests for the next cases

PR #807, opened 2024-02-24
Solution

yaroslavsysoiev:
Well, this function fails the test_invalid_input_types and does not raise any Exception, like ValueError. But I decided to leave this test.

### DIFF
--- a/app/test_main.py
+++ b/app/test_main.py
@@ -1,3 +1,35 @@
+from typing import Type
+import pytest
 from app.main import get_human_age
 
-# write your code here
+
+@pytest.mark.parametrize("input_age, next_age, expected_result", [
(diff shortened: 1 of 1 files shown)


app/test_main.py
```diff
@@ -1,3 +1,35 @@
...
+#     ("string", 25, KeyError),
+#     (20, "string", KeyError),
+# ])
+# def test_invalid_input_types(
+#         input_age: int,
+#         next_age: int,
+#         expected_error: Type[BaseException]
+# ) -> None:
+#     with pytest.raises(expected_error):
+#         get_human_age(input_age, next_age)
```
vsmutok, line 35: remove comments

vsmutok (changes requested):
image

PR #810, opened 2024-02-27
Solution

CreepyMo:


app/test_main.py
```diff
@@ -1,3 +1,43 @@
...
+        pytest.param(
+            27, 27, [2, 2],
+            id="27 cat/dog years should convert into 2 human age."
+        ),
+        pytest.param(
+            28, 28, [3, 2],
+            id="28 cat/dog years should convert into 3/2 human age."
+        )
+    ]
+)
```
Dimosphen1, line 37: Add tests that pass large numbers, negative numbers, and different numbers to `cat_age` and `dog_age` as mentioned in the checklist. Besides that, declare another function to test against the raised error in case different invalid data types are passed to the function

app/test_main.py
```diff
@@ -1,3 +1,43 @@
+import pytest
```
Dimosphen1, line 1: Add a blank line between different groups of imports according to PEP-8

Dimosphen1 (changes requested):

app/test_main.py
CreepyMo: Done

app/test_main.py
CreepyMo: Done

PR #826, opened 2024-03-11
Solution

mwellick:


EdAlekseiev (changes requested):
Add test case for the next case

app/test_main.py
```diff
@@ -1,3 +1,35 @@
...
+    (28, 28, [3, 2]),
+    (100, 100, [21, 17]),
+    (-5, 10, [0, 0]),
+    (5, -10, [0, 0]),
+    (-5, -10, [0, 0]),
+
+])
+def test_get_human_age(cat_age: int,
+                       dog_age: int,
+                       expected_result: list) -> None:
```
vasylhnatiuk, line 21:
```suggestion
                       expected_result: list[int]) -> None:
```

app/test_main.py
```diff
@@ -1,3 +1,35 @@
...
+    ("5", 10),
+    (5, "10"),
+    ("5", "10")
+])
+def test_get_human_age_raises_type_error(
+        age_in_human_years: int,
+        age_in_dog_years: int
+) -> None:
+    with pytest.raises(TypeError):
+        get_human_age("5", "10")
```
vasylhnatiuk, line 35:
```suggestion
        get_human_age(age_in_human_years, age_in_dog_years)
```

vasylhnatiuk: previous answer was correct, sorry

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #827, opened 2024-03-12
'Solution'

Meyertal:


YukihiroSM (changes requested):
Add some tests when function arguments are not integers (string, or None) -> the function should raise some exception. 

Also, test with really big numbers.

NikOneZ1 (approved):
Well done!

PR #830, opened 2024-03-13
Solution

ILyakhova:


NazarSheremet (changes requested):
In this task need to use `@pytest.mark.parametrize`
image

NazarSheremet (approved):
Good job!

PR #839, opened 2024-03-16
solution

kostiukmkalne:


NazarSheremet (changes requested):
image
Didn't implement this point

NazarSheremet (changes requested):
Good, but write more test. create one more `@pytest.mark.parametrize` in format:
```
@pytest.mark.parametrize(
    "cat_age, dog_age, error",
    [
    ]
 )
```

kostiukmkalne:
Can`t understand what is wrong?

app/test_main.py
```diff
@@ -1,3 +1,46 @@
...
+        get_human_age("10", 10)
+
+
+@pytest.mark.parametrize(
+    "cat_age, dog_age, expected_error",
+    [
+        (0, bool, TypeError),
+        (3.5, "cat", TypeError),
+        (10, 10.5, TypeError),
+        (10.5, 10.5, TypeError),
```
NazarSheremet, line 34: In these two cases, an error is expected, but it does not occur because the function supports float.

NazarSheremet (changes requested):

NazarSheremet (approved):
Good job!

PR #844, opened 2024-03-18
solution

SHCHERBANV:


NazarSheremet (changes requested):
image
Didn't implement this point

NazarSheremet (changes requested):
Good, but write more test. create one more @pytest.mark.parametrize in format:
```
@pytest.mark.parametrize(
    "cat_age, dog_age, error",
    [
    ]
 )
```

app/test_main.py
```diff
@@ -1,3 +1,69 @@
...
+
+@pytest.mark.parametrize(
+    "cat_age, dog_age, error",
+    [
+        (0, 0, False),
+        (15, 15, False),
+        (100, 100, False),
+        (200, 200, False),
+    ],
+)
```
NazarSheremet, line 57: Not quite so, you need to pass the expected error and data of such type to trigger it

```
 @pytest.mark.parametrize(
        "cat_age, dog_age, expected_error",
        [
            (0, bool, TypeError),
            (3.5, "cat", TypeError),
        ]
    )
```

NazarSheremet (changes requested):
Need changes

NazarSheremet (approved):
Good job!

PR #849, opened 2024-03-20
Solution

Bujhm99:


Boyok1337 (approved):

app/test_main.py
```diff
@@ -1,3 +1,42 @@
...
+                        dog_year: int,
+                        result: list) -> None:
+    func_result = get_human_age(cat_year, dog_year)
+    assert (func_result[0] == result[0]),\
+        (f"Somothing wrong with cat calculate: "
+         f"waited {result[0]}, get {func_result[0]}")
+
+    assert (func_result[1] == result[1]), \
+        (f"Somothing wrong with dog calculate: "
+         f"waited {result[1]}, get {func_result[1]}")
```
YuriiKindrat, line 42: why do u use indexes here ? could u check the whole lists ?
also thry to avoid using linebreakers here

app/test_main.py
```diff
@@ -1,3 +1,42 @@
...
+@pytest.mark.parametrize(
+    "cat_year, dog_year, result", [
+        (0, 0, [0, 0]),
+        (14, 14, [0, 0]),
+        (15, 15, [1, 1]),
+        (23, 23, [1, 1]),
+        (24, 24, [2, 2]),
+        (27, 27, [2, 2]),
+        (28, 28, [3, 2]),
+        (100, 100, [21, 17])
```
YuriiKindrat, line 29: it is a good practice to add ids to parametrize params

YuriiKindrat (changes requested):

app/test_main.py
```diff
@@ -1,3 +1,48 @@
...
 
-# write your code here
+
+@pytest.mark.parametrize(
+    "cat_year, dog_year, expect_error", [
+        pytest.param("5", 6.1, TypeError, id="cat value is not integer"),
+        pytest.param(5, "6", TypeError, id="dog value is not integer")
+    ]
+)
+def test_cheking_exeptions(cat_year: int,
```
Oleksl888, line 12: `checking`

app/test_main.py
```diff
@@ -1,3 +1,48 @@
...
+                           dog_year: int,
+                           expect_error: Exception) -> None:
```
Oleksl888, line 14: Here must be `Type[Exception]` to signify that you expect an exception as a class not an instance

app/test_main.py
```diff
@@ -1,3 +1,48 @@
...
+        "27 cat/dog years should convert into 2 human age.",
+        "28 year of cat = 3 human year, 28 dog years = 2 human age.",
+        "100 year of cat = 21 human year, 100 dog years = 17 human age."
+    ]
+)
+def test_cheking_values(cat_year: int,
+                        dog_year: int,
+                        result: list) -> None:
+    func_result = get_human_age(cat_year, dog_year)
+    print("func_result")
```
Oleksl888, line 45: What is this print here for?

app/test_main.py
```diff
@@ -1,3 +1,48 @@
...
+    assert (func_result
+            == result), (f"Somthing wrong with calculate: "
```
Oleksl888, line 47: Avoid grammar mistakes in error messages, also stylistically it would be better something like this :
`'x' cat years did not correctly convert to human years. Expected human years was 'y' but got 'z'.

Oleksl888 (changes requested):

YuriiKindrat (changes requested):

YuriiKindrat (approved):
GJ

PR #853, opened 2024-03-22
Solved

pochtalon:


NazarSheremet (changes requested):
In this task need to use `@pytest.mark.parametrize()` it will simplify your code 
Also didn't implement this point
image

app/main.py
```diff
@@ -1,4 +1,6 @@
 def get_human_age(cat_age: int, dog_age: int) -> list:
+    if cat_age < 0 or cat_age > 500 or dog_age < 0 or dog_age > 450:
+        raise ValueError("Invalid data")
```
NazarSheremet, line 3: No need to add this check and change main app

NazarSheremet (changes requested):

YuriiSmolii (approved):
LGTM

PR #871, opened 2024-03-31
Solution

panicua:


app/test_main.py
```diff
@@ -1,3 +1,55 @@
...
+            (23, 23, [1, 1]),
+            (24, 24, [2, 2]),
+            (100, 100, [21, 17])
+        ]
+    )
+    def test_should_return_age_normal_range_values(
+            self,
+            cat_age: int,
+            dog_age: int,
+            expected: int) -> None:
```
vitalinamal, line 20: here you are expected list[int]

app/test_main.py
```diff
@@ -1,3 +1,55 @@
...
+            (-1, -1, [0, 0]),
+            (500, 600, [121, 117]),
+            (2345272727234, 34563473457, [586318181804, 6912694688])
+        ],
+    )
+    def test_should_return_age_edge_range_values(
+            self,
+            cat_age: int,
+            dog_age: int,
+            expected: int) -> None:
```
vitalinamal, line 36: same here

app/test_main.py
```diff
@@ -1,3 +1,55 @@
...
+        assert get_human_age(cat_age, dog_age) == expected
+
+    @pytest.mark.parametrize(
+        "cat_age,dog_age",
```
vitalinamal, line 40: here you need to add what error you expect as the third argument.
this is how I did it, but maybe your option is correct, so you can ignore this comment)

app/test_main.py
```diff
@@ -1,3 +1,55 @@
...
+            ({"1": 1}, {"2": None}),
+            ({1}, {0}),
+            (None, None),
+        ]
+    )
+    def test_should_raise_error_if_age_not_int(
+            self,
+            cat_age: int,
+            dog_age: int) -> None:
+        with pytest.raises(TypeError):
```
vitalinamal, line 54: so you can use the third argument here instead of TypeError
but again, if I am correct )

vitalinamal (approved):
Good job)

app/test_main.py
panicua: > here you need to add what error you expect as the third argument. this is how I did it, but maybe your option is correct, so you can ignore this comment)

We only have one type of an error in our case. In this case I'll have to put everywhere 3rd argument ```TypeError``` instead of one inside of the function as I did.

app/test_main.py
panicua: same as I wrote above

app/test_main.py
```diff
@@ -1,3 +1,55 @@
...
+            (23, 23, [1, 1]),
+            (24, 24, [2, 2]),
+            (100, 100, [21, 17])
+        ]
+    )
+    def test_should_return_age_normal_range_values(
+            self,
+            cat_age: int,
+            dog_age: int,
+            expected: list) -> None:
```
YuriiKindrat, line 20: bad formatting here (

app/test_main.py
```diff
@@ -1,3 +1,55 @@
...
+            (500, 600, [121, 117]),
+            (2345272727234, 34563473457, [586318181804, 6912694688])
+        ],
+    )
+    def test_should_return_age_edge_range_values(
+            self,
+            cat_age: int,
+            dog_age: int,
+            expected: list) -> None:
+        assert get_human_age(cat_age, dog_age) == expected
```
YuriiKindrat, line 37: same is here

app/test_main.py
```diff
@@ -1,3 +1,55 @@
...
+            ([1], [1]),
+            ({"1": 1}, {"2": None}),
+            ({1}, {0}),
+            (None, None),
+        ]
+    )
+    def test_should_raise_error_if_age_not_int(
+            self,
+            cat_age: int,
+            dog_age: int) -> None:
```
YuriiKindrat, line 53: same

YuriiKindrat (changes requested):
it is common practice to ids to parametrize )

LiudmylaKulinchenko (approved):
Looks good! 🌷

PR #873, opened 2024-04-01
'Solution'

March1205:


app/test_main.py
```diff
@@ -1,3 +1,24 @@
...
+
+
+def test_get_human_age_24_to_27() -> None:
+    assert get_human_age(24, 24) == [2, 2]
+    assert get_human_age(27, 27) == [2, 2]
+
+
+def test_get_human_age_more_than_27() -> None:
+    assert get_human_age(28, 28) == [3, 2]
+    assert get_human_age(100, 100) == [21, 17]
```
vsmutok, line 24: image

vsmutok (changes requested):

vsmutok (changes requested):
image

EdAlekseiev (changes requested):
Add tests for the next case

PR #874, opened 2024-04-01
Solution

IlliaPetukhov:


EdAlekseiev (changes requested):
Try to use pytest parametrize
https://docs.pytest.org/en/7.1.x/how-to/parametrize.html

vsmutok (approved):
Looks Good!

PR #883, opened 2024-04-03
'Solution'

dkotkod:


app/test_main.py
```diff
@@ -1,3 +1,28 @@
...
+class TestHumanAgeConversion:
+    @pytest.mark.parametrize("cat_age, dog_age, expected", [
+        (0, 0, [0, 0]),
+        (14, 14, [0, 0]),
+        (15, 15, [1, 1]),
+        (23, 23, [1, 1]),
+        (24, 24, [2, 2]),
+        (27, 27, [2, 2]),
+        (28, 28, [3, 2]),
+        (100, 100, [21, 17]),
```
Oleksl888, line 14: > if function resive data out of normal range, such as negative numbers, zero, or realy large numbers;

this requirement is not covered 100%

app/test_main.py
```diff
@@ -1,3 +1,28 @@
+import pytest
 from app.main import get_human_age
 
-# write your code here
+
+class TestHumanAgeConversion:
+    @pytest.mark.parametrize("cat_age, dog_age, expected", [
+        (0, 0, [0, 0]),
```
Oleksl888, line 7: What about id/ids for data?

app/test_main.py
```diff
@@ -1,3 +1,28 @@
...
+            self,
+            cat_age: int,
+            dog_age: int,
+            expected: list[int]
+    ) -> None:
+        assert get_human_age(cat_age, dog_age) == expected
+
+    def test_non_integer_age(self) -> None:
+        with pytest.raises(TypeError):
+            get_human_age("cat", 10)
```
Oleksl888, line 26: This data should also be parametrized

Oleksl888 (changes requested):

app/test_main.py
```diff
@@ -1,3 +1,62 @@
...
+        (5, -1, [0, 0]),
+        (5, 0, [0, 0]),
+        (1000, 1000, [246, 197]),
+    ], ids=[
+        "Both zero age",
+        "Both small age",
+        "Both middle age",
+        "Both middle age",
+        "Both middle age",
+        "Both middle age",
```
Oleksl888, line 26: Add more descriptive comments such as `both cats and dogs years equals 1 human year` etc

app/test_main.py
```diff
@@ -1,3 +1,62 @@
...
+    ) -> None:
+        assert get_human_age(cat_age, dog_age) == expected
+
+    @pytest.mark.parametrize("invalid_cat_age, invalid_dog_age", [
+        ("cat", 10),
+        (5, "dog"),
+        (5.5, 10),
+        (5, 10.5),
+        (None, 10),
+        (5, None)
```
Oleksl888, line 49: what about ids here? Make it meaningful also

app/test_main.py
```diff
@@ -1,3 +1,62 @@
...
+    ])
+    def test_non_integer_age_invalid_type(
+            self,
+            invalid_cat_age: int,
+            invalid_dog_age: int
+    ) -> None:
+        with pytest.raises(TypeError):
+            if (
```
Oleksl888, line 57: no need to add extra conditions and raise error inside of a test that checks for raising error

app/test_main.py
```diff
@@ -1,3 +1,62 @@
...
+        ("cat", 10),
+        (5, "dog"),
+        (5.5, 10),
+        (5, 10.5),
+        (None, 10),
+        (5, None)
+    ])
+    def test_non_integer_age_invalid_type(
+            self,
+            invalid_cat_age: int,
```
Oleksl888, line 53: invalid params will be of different types so it is appropriate to use `Any` here

Oleksl888 (changes requested):

app/test_main.py
dkotkod: if I remove this condition, then 2 tests fail

LiudmylaKulinchenko (changes requested):

app/test_main.py
```diff
@@ -1,3 +1,61 @@
+import pytest
+from typing import Any
```
Oleksl888, line 2: order of imports is incorrect here, also missing blank lines

Oleksl888 (changes requested):

app/test_main.py
```diff
@@ -1,3 +1,72 @@
...
+            cat_age: int,
+            dog_age: int,
+            expected: list[int]
+    ) -> None:
+        assert get_human_age(cat_age, dog_age) == expected
+
+    @pytest.mark.parametrize("invalid_cat_age, invalid_dog_age", [
+        ("cat", 10),
+        (5, "dog"),
+        (5.5, 10),
```
Oleksl888, line 49: this condition will not produce typeerror

app/test_main.py
Oleksl888: Please remove if conditions and raise of TypeError here. 
It is unnecessary since the purpose of this test is to check if the function that we test will produce an exception if we enter incorrect data. Therefor if you put only incorrect data in params your test will pass with only 
```
with pytest.raises(TypeError):
   func()

Oleksl888 (changes requested):

PR #884, opened 2024-04-03
solution

alonasorochynska:


app/test_main.py
```diff
@@ -1,3 +1,34 @@
 from app.main import get_human_age
 
-# write your code here
+import pytest
```
Oleksl888, line 3: order of imports?

app/test_main.py
```diff
@@ -1,3 +1,34 @@
...
+                     id="The next 9 pets years give 1 more human year"),
+        pytest.param(27, 27, [2, 2],
+                     id="The next 9 pets years give 1 more human year"),
+        pytest.param(28, 28, [3, 2],
+                     id="Every 4 next cat years give 1 extra human year"),
+        pytest.param(100, 100, [21, 17],
+                     id="Every 4 next cat years give 1 extra human year"),
+    ]
+)
+def test_correct_math(
```
Oleksl888, line 29: Choose more relevant name for this test

app/test_main.py
```diff
@@ -1,3 +1,34 @@
 from app.main import get_human_age
 
-# write your code here
+import pytest
+
+
+@pytest.mark.parametrize(
+    "cat_age, dog_age, expected_result",
```
Oleksl888, line 7: Please refer to the checklist p2 and add missing requirements

Oleksl888 (changes requested):

app/test_main.py
```diff
@@ -1,3 +1,57 @@
...
+    [
+        pytest.param("-1", "-1",
+                     id="Ages have to be integers"),
+        pytest.param("0", 0,
+                     id="Cats age have to be an integer"),
+        pytest.param(0, "0",
+                     id="Dogs age have to be an integer"),
+    ]
+)
+def test_types_of_input_data(cat_age: int, dog_age: int) -> None:
```
Oleksl888, line 49: `int` is not a valid type here since we testing for different inputs.
Also you could add few more test params with data other then str and int

Oleksl888 (changes requested):

Oleksl888 (approved):
NIce!

PR #885, opened 2024-04-03
Solution

Nyxan:


app/test_main.py
```diff
@@ -1,3 +1,28 @@
...
 from app.main import get_human_age
 
-# write your code here
+
+@pytest.mark.parametrize(
+    "cat_age, dog_age, expected",
+    [
+        (0, 0, [0, 0]),
+        (14, 14, [0, 0]),
+        (24, 24, [2, 2]),
```
Oleksl888, line 10: Must add more cases here:

> if output should change with some integer value you should check that it is not changed with previous value;

check p2 from checklist so you cover everything

app/test_main.py
```diff
@@ -1,3 +1,28 @@
...
+         "Dog age greater than cat age",
+         "Large numbers", "Negative ages"]
+)
+def test_get_human_age(cat_age: int, dog_age: int, expected: list) -> None:
+    assert get_human_age(cat_age, dog_age) == expected
+
+
+def test_get_human_age_type_error() -> None:
+    with pytest.raises(TypeError):
+        get_human_age("cat", "dog")
```
Oleksl888, line 28: parametrize input and add more data types combinations to check for error

Oleksl888 (changes requested):

app/test_main.py
```diff
@@ -1,3 +1,45 @@
...
+    [
+        ("cat", "dog"),
+        (None, None),
+        ({}, []),
+    ],
+    ids=["String inputs",
+         "None inputs",
+         "Unsupported types"]
+)
+def test_get_human_age_type_error(cat_age: any, dog_age: any) -> None:
```
Oleksl888, line 43: `any` is a built in function, it is not a valid type hint

app/test_main.py
```diff
@@ -1,3 +1,45 @@
...
+    ],
+    ids=["Both ages are zero",
+         "Both ages are less than the first_year",
+         "Both ages are greater than the second_year",
+         "Dog input greater, ages are equal",
+         "Cat input greater, ages are equal",
+         "Ages are equal",
+         "Different ages between dogs and cats"
+         "Dog age greater than cat age",
+         "Large numbers", "Negative ages"]
```
Oleksl888, line 26: last element should be on a separate line

app/test_main.py
```diff
@@ -1,3 +1,45 @@
...
+
+
+@pytest.mark.parametrize(
+    "cat_age, dog_age",
+    [
+        ("cat", "dog"),
+        (None, None),
+        ({}, []),
+    ],
+    ids=["String inputs",
```
Oleksl888, line 39: think of better names such as `TypeError raised when input is a string`

Oleksl888 (changes requested):

LiudmylaKulinchenko (approved):
Good job ✨

PR #886, opened 2024-04-03
Solution for task

oleksandra-shershen:


app/test_main.py
```diff
@@ -1,3 +1,33 @@
...
+    "cat_age, dog_age, expected_human_ages",
+    [
+        (0, 0, [0, 0]),
+        (14, 14, [0, 0]),
+        (15, 15, [1, 1]),
+        (23, 23, [1, 1]),
+        (24, 24, [2, 2]),
+        (27, 27, [2, 2]),
+        (28, 28, [3, 2]),
+        (100, 100, [21, 17]),
```
Oleksl888, line 15: All requirements are not fulfilled - please refer to checkilst p2

Oleksl888 (changes requested):

app/test_main.py
oleksandra-shershen: Resolved

Oleksl888: Almost:)
 what about negative numbers?

app/test_main.py
```diff
@@ -1,3 +1,56 @@
+import pytest
+from typing import Any, Type
```
Oleksl888, line 2: order of import here?

app/test_main.py
```diff
@@ -1,3 +1,56 @@
...
+    ids=[
+        "None should raise TypeError",
+        "str should raise TypeError",
+        "set should raise TypeError",
+    ]
+)
+def test_test_get_human_age_errors(
+        cat_age: Any,
+        dog_age: Any,
+        expected_error: Type[BaseException]
```
Oleksl888, line 53: Here should be actually TypeError since you defined only this exception in parametrizer

app/test_main.py
oleksandra-shershen: Resolved

app/test_main.py
oleksandra-shershen: Resolved

app/test_main.py
```diff
@@ -1,3 +1,62 @@
+from typing import Any, Type
+import pytest
```
Oleksl888, line 2: add blank lines between imports

app/test_main.py
```diff
@@ -1,3 +1,62 @@
...
+        dog_age: int,
+        expected_human_ages: list[int]
+) -> None:
+    assert get_human_age(cat_age, dog_age) == expected_human_ages
+
+
+@pytest.mark.parametrize(
+    "cat_age, dog_age, expected_error",
+    [
+        (None, None, TypeError),
```
Oleksl888, line 46: no need to parametrize TypeError since you have only one type of error

app/test_main.py
```diff
@@ -1,3 +1,62 @@
...
+        ("10", "10", TypeError),
+        ({1, 2}, {1, 2}, TypeError),
+    ],
+    ids=[
+        "None should raise TypeError",
+        "str should raise TypeError",
+        "set should raise TypeError",
+    ]
+)
+def test_test_get_human_age_errors(
```
Oleksl888, line 56: think of a better name for this test?

Oleksl888 (changes requested):

Oleksl888 (approved):
Great!

PR #888, opened 2024-04-03
Solution 1

semermaks:


app/test_main.py
```diff
@@ -1,3 +1,28 @@
...
+
+class TestHumanAgeConversion:
+    @pytest.mark.parametrize("cat_age, dog_age, expected", [
+        (0, 0, [0, 0]),
+        (14, 14, [0, 0]),
+        (15, 15, [1, 1]),
+        (23, 23, [1, 1]),
+        (24, 24, [2, 2]),
+        (27, 27, [2, 2]),
+        (28, 28, [3, 2]),
```
Oleksl888, line 13: not all requirements from checklist are fulfilled. Check p2

app/test_main.py
```diff
@@ -1,3 +1,28 @@
...
+            dog_age: int,
+            expected: list[int]
+    ) -> None:
+        assert get_human_age(cat_age, dog_age) == expected
+
+    def test_non_integer_age(self) -> None:
+        with pytest.raises(TypeError):
+            get_human_age("cat", 10)
+        with pytest.raises(TypeError):
+            get_human_age(5, "dog")
```
Oleksl888, line 28: you should parametrize this + add different data types that may cause an error. You have str + int and int + str which is basically same

Oleksl888 (changes requested):

app/test_main.py
```diff
@@ -1,3 +1,44 @@
...
+                         id="`None` should raise `TypeError`"),
+            pytest.param("10", "10", TypeError,
+                         id="`str` should raise `TypeError`"),
+            pytest.param({1, 2}, {1, 2}, TypeError,
+                         id="`set` should raise `TypeError`")
+        ]
+    )
+    def test_test_get_human_age_errors(
+            self,
+            cat_age: any,
```
Oleksl888, line 39: `any` is a built in function - not a valid type hint

app/test_main.py
```diff
@@ -1,3 +1,44 @@
...
+    def test_get_human_age(self,
+                           cat_age: int,
+                           dog_age: int,
+                           expected_result: list[int]) -> None:
+        assert get_human_age(cat_age, dog_age) == expected_result
+
+    @pytest.mark.parametrize(
+        "cat_age, dog_age, expected_error",
+        [
+            pytest.param(None, None, TypeError,
```
Oleksl888, line 29: No need to add TypeError to params since you will use only one type of error here

app/test_main.py
```diff
@@ -1,3 +1,43 @@
+import pytest
```
Oleksl888, line 1: different import types must be separated

app/test_main.py
```diff
@@ -1,3 +1,43 @@
...
+        [
+            pytest.param(-1, -1, [0, 0]),
+            pytest.param(0, 0, [0, 0]),
+            pytest.param(14, 14, [0, 0]),
+            pytest.param(15, 15, [1, 1]),
+            pytest.param(23, 23, [1, 1]),
+            pytest.param(24, 24, [2, 2]),
+            pytest.param(27, 27, [2, 2]),
+            pytest.param(28, 28, [3, 2]),
+            pytest.param(100, 100, [21, 17]),
```
Oleksl888, line 17: what about id/ids here?

app/test_main.py
```diff
@@ -1,3 +1,43 @@
...
+                         id="`None` should raise `TypeError`"),
+            pytest.param("10", "10",
+                         id="`str` should raise `TypeError`"),
+            pytest.param({1, 2}, {1, 2},
+                         id="`set` should raise `TypeError`")
+        ]
+    )
+    def test_test_get_human_age_errors(
+            self,
+            cat_age: int,
```
Oleksl888, line 39: these types are not ints in this test

Oleksl888 (changes requested):

app/test_main.py
semermaks: What type are they?😅

semermaks: What type are they?))

Oleksl888: Any from typing

app/test_main.py
Oleksl888: Check order of imports + blank lines must be added

app/test_main.py
Oleksl888: not done

Oleksl888 (changes requested):

LiudmylaKulinchenko (approved):
Good job 😌

PR #890, opened 2024-04-03
'Solution'

kssshevchenko:


YuriiKindrat (changes requested):
image
fix that 
also it is common practice to add ids to ur parametrize

PR #891, opened 2024-04-03
My tests

Chub-andrew:


app/test_main.py
```diff
@@ -1,3 +1,32 @@
...
+    (14, 14, [0, 0]),
+    (15, 15, [1, 1]),
+    (23, 23, [1, 1]),
+    (24, 24, [2, 2]),
+    (27, 27, [2, 2]),
+    (28, 28, [3, 2])])
+def test_get_human_age_with_param_deco(
+        cat_age: int,
+        dog_age: int,
+        expected_result: list) -> None:
```
YuriiKindrat, line 17: bad formatting here

app/test_main.py
```diff
@@ -1,3 +1,32 @@
...
-# write your code here
+
+@pytest.mark.parametrize("cat_age, dog_age, expected_result", [
+    (0, 0, [0, 0]),
+    (14, 14, [0, 0]),
+    (15, 15, [1, 1]),
+    (23, 23, [1, 1]),
+    (24, 24, [2, 2]),
+    (27, 27, [2, 2]),
+    (28, 28, [3, 2])])
```
YuriiKindrat, line 13: what about big/negative numbers ?

YuriiKindrat (changes requested):

app/test_main.py
```diff
@@ -1,3 +1,34 @@
...
+    (0, 0, [0, 0]),
+    (14, 14, [0, 0]),
+    (15, 15, [1, 1]),
+    (23, 23, [1, 1]),
+    (24, 24, [2, 2]),
+    (27, 27, [2, 2]),
+    (28, 28, [3, 2]),
+    (-1, -1, [0, 0]),
+    (100, 100, [21, 17])
+])
```
Oleksl888, line 16: what about ids here and in below test?

app/test_main.py
```diff
@@ -1,3 +1,34 @@
...
+def test_get_human_age_with_param_deco(cat_age: int,
+                                       dog_age: int,
+                                       expected_result: list) -> None:
+    if isinstance(expected_result, list):
```
Oleksl888, line 20: Since you are the one that writes this test and explicitly puts list in parameters why do you need this check here?

app/test_main.py
```diff
@@ -1,3 +1,34 @@
...
+        result = get_human_age(cat_age, dog_age)
```
Oleksl888, line 21: no need to store result into variable since you use it only once

Oleksl888 (changes requested):

app/test_main.py
```diff
@@ -1,3 +1,35 @@
...
+    (27, 27, [2, 2]),
+    (28, 28, [3, 2]),
+    (-1, -1, [0, 0]),
+    (100, 100, [21, 17])
+])
+def test_get_human_age_with_param_deco(cat_age: int,
+                                       dog_age: int,
+                                       expected_result: list) -> None:
+    assert get_human_age(cat_age, dog_age) == expected_result, (
+        f"Function 'get_human_age' should return {expected_result}"
```
Oleksl888, line 21: You must write description like this for every param in parameterizer `id or ids` attribute above. Such as:
`id="expected human age 0 for cat age 0 and dog age 0"`

app/test_main.py
```diff
@@ -1,3 +1,35 @@
...
+        f"when value is Cat: {cat_age}, Dog: {dog_age}"
+    )
+
+
+@pytest.mark.parametrize("cat_age, dog_age, expected_errors", [
+    ([], [], TypeError),
+    ((), (), TypeError),
+    ("", "", TypeError),
+    ({}, {}, TypeError)])
```
Oleksl888, line 30: Same here you need to add meaningful description for dataset that you are testing:
`TypeError raised when input is two lists`

app/test_main.py
```diff
@@ -1,3 +1,35 @@
...
+def test_get_human_age_with_param_error(cat_age: int,
+                                        dog_age: int,
+                                        expected_errors: list) -> None:
```
Oleksl888, line 33: expected errors is not a list type. Also since you test only one type of error you can simply pass it like so: `pytest.raises(TypeError)`

Oleksl888 (changes requested):

app/main.py
```diff
@@ -1,4 +1,8 @@
 def get_human_age(cat_age: int, dog_age: int) -> list:
+    """
```
Oleksl888, line 2: why make changes to the original function?

app/test_main.py
```diff
@@ -1,3 +1,46 @@
...
+        "Expected human age 2 for cat age 24 and dog age 24",
+        "Expected human age 2 for cat age 27 and dog age 27",
+        "Expected human age for cat:3 and dog:2 for cat age 28 and dog age 28",
+        "Expected human age 0 for cat age -1 and dog age -1",
+        "Expected human age for cat:21 and dog:17 for cat and dog ages 100"])
+def test_get_human_age_with_param_deco(cat_age: int,
+                                       dog_age: int,
+                                       expected_result: list) -> None:
+    assert get_human_age(cat_age, dog_age) == expected_result, (
+        f"Function 'get_human_age' should return {expected_result}"
```
Oleksl888, line 29: you don't need an error message in assert since you already defined the ids

app/test_main.py
```diff
@@ -1,3 +1,46 @@
...
+        f"when value is Cat: {cat_age}, Dog: {dog_age}"
+    )
+
+
+@pytest.mark.parametrize("cat_age, dog_age, expected_errors", [
+    ([], [], TypeError),
```
Oleksl888, line 35: why you parametrize error if you don't use it in your test anyway?

app/test_main.py
```diff
@@ -1,3 +1,46 @@
...
+    ((), (), TypeError),
+    ("", "", TypeError),
+    ({}, {}, TypeError)], ids=["TypeError raised when input is two lists",
```
Oleksl888, line 38: it is better to put long params like these each on separate line
```
({}, {}, TypeError)], 
ids=[
    "TypeError raised when input is two lists",
   ...

**Review by Oleksl888 — CHANGES_REQUESTED** [review]

File: app/test_main.py [hunk 13]
```diff
@@ -1,3 +1,44 @@
...
+    ((), ()),
+    ("", ""),
+    ({}, {})],
+    ids=["TypeError raised when input is two lists",
+         "TypeError raised when input is two tuples",
+         "TypeError raised when input is two strings",
+         "TypeError raised when input is two dicts"])
+def test_get_human_age_with_param_error(cat_age: int,
+                                        dog_age: int,
+                                        ) -> None:
```
Oleksl888, line 42: bad formatting here

app/test_main.py
```diff
@@ -1,3 +1,44 @@
...
+@pytest.mark.parametrize("cat_age, dog_age", [
+    ([], []),
+    ((), ()),
+    ("", ""),
+    ({}, {})],
+    ids=["TypeError raised when input is two lists",
+         "TypeError raised when input is two tuples",
+         "TypeError raised when input is two strings",
+         "TypeError raised when input is two dicts"])
+def test_get_human_age_with_param_error(cat_age: int,
```
Oleksl888, line 40: in this test we expect not ints but any possible types

app/main.py
```diff
@@ -1,4 +1,8 @@
...
+
+    :rtype: object
```
Oleksl888, line 4: please revert this change

Oleksl888 (changes requested):

YuriiKindrat (approved):
GJ

PR #892, opened 2024-04-03
Solution#1

VitalyBashkiser:


YuriiKindrat (changes requested):
image
fix that

app/test_main.py
```diff
@@ -1,3 +1,47 @@
...
+        "cat_age, dog_age, expected",
+        [
+            (0, 0, [0, 0]),
+            (14, 14, [0, 0]),
+            (15, 15, [1, 1]),
+            (23, 23, [1, 1]),
+            (24, 24, [2, 2]),
+            (27, 27, [2, 2]),
+            (28, 28, [3, 2]),
+            (100, 100, [21, 17]),
```
Oleksl888, line 18: You should include negative parameters in this test

app/test_main.py
```diff
@@ -1,3 +1,47 @@
...
-# write your code here
+
+class TestGetHumanAge:
+    # Basic test cases
+    @pytest.mark.parametrize(
+        "cat_age, dog_age, expected",
+        [
+            (0, 0, [0, 0]),
+            (14, 14, [0, 0]),
+            (15, 15, [1, 1]),
```
Oleksl888, line 13: Also you should add meaningful description to the parameters

app/test_main.py
```diff
@@ -1,3 +1,47 @@
+import pytest
+from typing import List
```
Oleksl888, line 2: check order of imports here. What should come first?

app/test_main.py
```diff
@@ -1,3 +1,47 @@
...
+    def test_basic_cases(
+            self,
+            cat_age: int,
+            dog_age: int,
+            expected: List[int]
+    ) -> None:
+        assert get_human_age(cat_age, dog_age) == expected
+
+
+class TestEdgeCases:
```
Oleksl888, line 30: no need to create an extra class for this - you can include this method in above class

app/test_main.py
```diff
@@ -1,3 +1,47 @@
...
+            ("0", 10),   # Zero cat age
+            (10, "0"),   # Zero dog age
+            ("1000", "1000"),  # Very large ages
+            ("10", 10),  # String cat age
+            (10, "10"),  # String dog age
+            ([10], 10),  # List cat age
+            (10, [10]),  # List dog age
+        ]
+    )
+    def test_edge_cases(self, cat_age: int, dog_age: int) -> None:
```
Oleksl888, line 45: edge cases are parameters that will not cause error but produce unrealistic result: such as a cat that is 1000 years old. Edge cases should be included in above test. Here you keep only params that will produce an error in the function - so you must rename this test to something like `test_error_will_be raised`

app/test_main.py
```diff
@@ -1,3 +1,47 @@
...
+    ) -> None:
+        assert get_human_age(cat_age, dog_age) == expected
+
+
+class TestEdgeCases:
+    @pytest.mark.parametrize(
+        "cat_age, dog_age",
+        [
+            ("-1", 10),  # Negative cat age
+            (10, "1"),  # Negative dog age
```
Oleksl888, line 35: remove comments and add proper description with `id`

app/test_main.py
Oleksl888, line 45: cat_age and dog_age will not be of type int here

Oleksl888 (changes requested):

app/test_main.py
```diff
@@ -1,3 +1,74 @@
+from typing import List, Any
+import pytest
```
Oleksl888, line 2: there should be blank lines between different types of imports

app/test_main.py
```diff
@@ -1,3 +1,74 @@
...
 from app.main import get_human_age
 
-# write your code here
+
+class TestGetHumanAge:
+    # Basic test cases
```
Oleksl888, line 7: remove unnecessary comments

app/test_main.py
```diff
@@ -1,3 +1,74 @@
...
+            "Boundary ages for young",
+            "Age just above the boundary",
+            "Boundary ages for adult",
+            "Age just above the boundary",
+            "Boundary ages for senior cat",
+        ]
+    )
+    def test_basic_cases(
+            self,
+            cat_age: Any,
```
Oleksl888, line 37: in this test cat_age and dog_age are always int type, expected is always a list of ints

app/test_main.py
```diff
@@ -1,3 +1,74 @@
...
+            "Negative cat age",
+            "Negative dog age",
+            "Zero ages",
+            "Both animals are in young age",
+            "Both animals are in young age, just above the boundary",
+            "Boundary ages for young",
+            "Age just above the boundary",
+            "Boundary ages for adult",
+            "Age just above the boundary",
+            "Boundary ages for senior cat",
```
Oleksl888, line 32: Most of these description do not represent what we are testing. Must be something like '100 years old cat and dog results 21 and 17 human ages

app/test_main.py
```diff
@@ -1,3 +1,74 @@
...
+            (10, "0"),
+            ("1000", "1000"),
+            ("10", 10),
+            (10, "10"),
+            ([10], 10),
+            (10, [10]),
+        ],
+        ids=[
+            "Negative cat age",
+            "Negative dog age",
```
Oleksl888, line 58: same here - it doesn't really represent what we are testing for. Example: 'TypeError raised when dog age is a string'

Oleksl888 (changes requested):

app/test_main.py
```diff
@@ -1,3 +1,75 @@
...
+            ("0", 10),
+            (10, "0"),
+            ("1000", "1000"),
+            ("10", 10),
+            (10, "10"),
+            ([10], 10),
+            (10, [10]),
+        ],
+        ids=[
+            "Cat age is a string",
```
Oleksl888, line 58: These descriptions should make it easier for  another person to understand what your test is about. It must have a meaningful name - such as `TypeError raised when dog age is a string`. Hope you follow this advice in future projects

Oleksl888 (approved):
Approved

PR #899, opened 2024-04-06
Solution

Sergunshot:


EdAlekseiev (changes requested):
Use pytest.parametrize for this task, please

vasylhnatiuk (changes requested):
image
do it

EdAlekseiev (changes requested):
Please, add test for the next case

EdAlekseiev (approved):

PR #901, opened 2024-04-07
Solution

VladyslavKsh:


app/test_main.py
```diff
@@ -1,3 +1,51 @@
...
+                28,
+                28,
+                [3, 2],
+                id="must be 3 years for cat age and 2 years for dog age"
+            ),
+            pytest.param(
+                65,
+                65,
+                [12, 10],
+                id="must return the correct age for every next years"
```
Oleksl888, line 43: not all edge cases covered, error handling not covered - please fix as per the task checklist

app/test_main.py
```diff
@@ -1,3 +1,51 @@
+import pytest
```
Oleksl888, line 1: must be blank line between different import types

Oleksl888 (changes requested):

app/test_main.py
VladyslavKsh: By error handling, did you mean incorrect values ​​at the function input? For example negative values ​​or data type error? If so, is it possible to change the code in the main file, for example, to add a raise condition, or should this only be done in tests?

Oleksl888: negative numbers will not produce error - you must include negative and really large numbers in these params as per checklist. Errors may be triggered only by wrong data types

app/test_main.py
```diff
@@ -1,3 +1,71 @@
...
+            )
+        ]
+    )
+    def test_get_correct_human_age(self,
+                                   cat_age: int,
+                                   dog_age: int,
+                                   expected_age_values: list) -> None:
+        assert get_human_age(cat_age, dog_age) == expected_age_values
+
+    def test_invalid_input(self) -> None:
```
Oleksl888, line 66: you must parametrize this with diffrerent data type combinations. pytest.raises() must be called only once

Oleksl888 (changes requested):

app/test_main.py
VladyslavKsh: > negative numbers will not produce error - you must include negative and really large numbers in these params as per checklist. Errors may be triggered only by wrong data types

I see, so there is no problem with negative numbers, because the function doesn't have a condition that the numbers must be greater than zero. So, can I add such test cases and write the result to expected, even if it is incorrect (value of age less than zero)?

YuriiKindrat (approved):
GJ

PR #902, opened 2024-04-07
Solution

siaserb:


app/test_main.py
```diff
@@ -1,3 +1,29 @@
...
+                                        "should be equal to 2"),
+        pytest.param(27, 27, [2, 2], id="output of this limit value "
+                                        "for cat equals 2 and for dog too"),
+        pytest.param(28, 28, [3, 2], id="output of this limit value "
+                                        "for cat equals 3 but for dog 2"),
+        pytest.param(29, 29, [3, 3], id="output of this limit value "
+                                        "for cat equals 3 and for dog too"),
+        pytest.param(100, 100, [21, 17], id="checking big values")
+    ]
+)
```
Oleksl888, line 27: missing some checklist requirements: handling negative numbers and raising errors

Oleksl888 (changes requested):

app/test_main.py
siaserb: But do I have the right to change the two main functions from the main.py?  Because no exceptions are raised there, and accordingly the test will always fail.

Oleksl888: You don’t have to. Just make test where you will pass incorrect parameters and test for exception being raised

siaserb: Due to the fact that the main function does not pass this test, Github shows me that all checks have failed

siaserb:
```
@pytest.mark.parametrize(
    "cat_age,dog_age,expected_error",
    [
        pytest.param(-2,
                     -36,
                     ValueError,
                     id="function should raise error if input is negative"),

        pytest.param(987654321,
                     1000000,
                     ValueError,
                     id="function should raise "
                        "error if input is large numbers")
    ]
)
def test_get_human_age_errors(cat_age: int,
                              dog_age: int,
                              expected_error: Exception) -> None:

```

Oleksl888: There will be no valueerror here because you are passing two ints. the function may be cause TypeError if one of the arguments will be not a numeric type

siaserb: But what type of exception should I use here? If ValueError is incorrect and TypeError for values which are not numeric and I have only numeric input...

Oleksl888: You can pass strings as params and expect TypeError out of it for example

siaserb: I'm so sorry. But how I should check negative numbers?

Oleksl888: you've done it correct - neg numbers will return 0,0 and no error

app/test_main.py
```diff
@@ -1,3 +1,80 @@
...
+def test_get_human_age(cat_age: int, dog_age: int, result: list[int]) -> None:
+    assert get_human_age(cat_age, dog_age) == result
+
+
+@pytest.mark.parametrize(
+    "cat_age,dog_age,expected_error",
+    [
+        pytest.param("test",
+                     "test",
+                     TypeError,
```
Oleksl888, line 72: you could add more params here to test different data types. Also no need to parametrize error type since it always will be TypeError

Oleksl888 (approved):
Approved. But please check and fix last comments

PR #905, opened 2024-04-08
Solution

OttoLindholm:


app/test_main.py
```diff
@@ -1,3 +1,21 @@
...
+def test_cat_and_dog_next_9_years() -> None:
+    assert get_human_age(24, 24) == [2, 2]
+
+
+def test_every_4_cat_years() -> None:
+    assert get_human_age(28, 28) == [3, 2]
+
+
+def test_every_5_dog_years() -> None:
+    assert get_human_age(100, 100) == [21, 17]
```
EdAlekseiev, line 21: Use pytest.mark.parametrize here

EdAlekseiev (changes requested):

EdAlekseiev (changes requested):
Add tests for the next cases

EdAlekseiev (approved):

PR #911, opened 2024-04-11
solution

ArtemLeo:


app/test_main.py
```diff
@@ -1,3 +1,26 @@
...
+        [
+            (-1, -1, [0, 0]),
+            (0, 0, [0, 0]),
+            (14, 14, [0, 0]),
+            (15, 15, [1, 1]),
+            (23, 23, [1, 1]),
+            (24, 24, [2, 2]),
+            (27, 27, [2, 2]),
+            (28, 28, [3, 2]),
+            (100, 100, [21, 17]),
```
Dimosphen1, line 17: Add test case to test against different values for `cat_age` and `dog_age` as mentioned in the checklist

app/test_main.py
```diff
@@ -1,3 +1,26 @@
...
+        ]
+    )
+    def test_get_human_age_returns_correct_age_for_cat_and_dog(
+            self,
+            cat_age: int,
+            dog_age: int,
+            expected_result: list
+    ) -> None:
+        assert get_human_age(cat_age, dog_age) == expected_result
```
Dimosphen1, line 26: Add test cases to test against invalid inputs and raised errors, when string, dict, list, etc. is passed to the function. This test is mentioned in the checklist as well

app/test_main.py
```diff
@@ -1,3 +1,26 @@
+import pytest
```
Dimosphen1, line 1: Add a blank line between different import groups to follow PEP-8

Dimosphen1 (changes requested):

app/test_main.py
ArtemLeo: added test

app/test_main.py
ArtemLeo: added test

app/test_main.py
ArtemLeo: added a blank

app/test_main.py
```diff
@@ -1,3 +1,48 @@
...
+            (-1, -1, [0, 0]),
+            (0, 0, [0, 0]),
+            (14, 14, [0, 0]),
+            (15, 15, [1, 1]),
+            (23, 23, [1, 1]),
+            (24, 24, [2, 2]),
+            (27, 27, [2, 2]),
+            (28, 28, [3, 2]),
+            (100, 100, [21, 17]),
+            (5, 10, [0, 0]),
```
Dimosphen1, line 19: Add test cases that would be variable enough in the future, so that `expected_result` would have a pair of values other than `0`

Dimosphen1 (approved):
LGTM, one comment is left for consideration

PR #919, opened 2024-04-15
'Solution'

Sonemon:


EdAlekseiev (changes requested):
Add tests for the next cases

Sonemon:
> Add tests for the next cases
> 
> * if function resive data out of normal range, such as negative numbers, zero, or realy large numbers;
> * if the function receives an incorrect type of data it raises the correct exception.

Do these tests need to be written for convert_to_human? Should they be positive tests for negative and large numbers? And negative tests for wrong type? And how do I test for realy large numbers, a bit confused how we define the boundary here?
Thanks.

Sonemon:
> Add tests for the next cases
> 
> * if function resive data out of normal range, such as negative numbers, zero, or realy large numbers;
> * if the function receives an incorrect type of data it raises the correct exception.

Sonemon:
> Add tests for the next cases
> 
> * if function resive data out of normal range, such as negative numbers, zero, or realy large numbers;
> * if the function receives an incorrect type of data it raises the correct exception.

I don't understand the assignment much: 
1. Negative values I added.
2. zero values have already been checked in the first variant
3. the values 100, 100 are the limit values for the life of cats and dogs or what kind of large values are needed?
4. I don't understand how to search all possible TypeError for values. Is it possible to check values for cat/dog separately via type(x) == int? Or in some other way?

EdAlekseiev (approved):

EdAlekseiev:
> > Add tests for the next cases
> > 
> > * if function resive data out of normal range, such as negative numbers, zero, or realy large numbers;
> > * if the function receives an incorrect type of data it raises the correct exception.
> 
> I don't understand the assignment much:
> 
> 1. Negative values I added.
> 2. zero values have already been checked in the first variant
> 3. the values 100, 100 are the limit values for the life of cats and dogs or what kind of large values are needed?
> 4. I don't understand how to search all possible TypeError for values. Is it possible to check values for cat/dog separately via type(x) == int? Or in some other way?

1. It's fine now
2. It's fine now
3. You should add test case for any large number. 100 is enough
4. Now, you test case TypeError is fine. You shouldn't test all possible TypeError for values, just a few of them

PR #923, opened 2024-04-17
Solution

fishray33:


fishray33:
I disagree with first check fail, just noticed the function with params (24.5, 30.5) returns not int type of list. So that what I tried to chek also. Or float is allowed to be as return type of list?

app/test_main.py
```diff
@@ -1,3 +1,46 @@
...
+    ]
+)
+def test_if_decimal_years_discard_remainder(
+        cat_age: int, dog_age: int, human_age_list: list[int]
+) -> None:
+    assert get_human_age(cat_age, dog_age) == human_age_list, \
+        "decimal part in year value should be discarded"
+    # assert (isinstance(get_human_age(cat_age, dog_age)[0], int)
+    #         and isinstance(get_human_age(cat_age, dog_age)[1], int)), \
+    #     "list of animals age should be int-type"
```
YuriiSmolii, line 26: remove commented code.

YuriiSmolii (changes requested):
Yes, function returns float values by default.
image

Clean your code.

app/test_main.py
fishray33: ok, thanx

YuriiSmolii (approved):
LGTM

PR #926, opened 2024-04-18
Solution

salaxambar:


YuriiSmolii (changes requested):
You should use mock to complete this task.

PR #940, opened 2024-05-07
solution

NAZ-KOM:


app/test_main.py
```diff
@@ -1,3 +1,13 @@
...
+def test_cat_dog_age_must_be_0_when_under_15() -> None:
+    assert get_human_age(13, 13) == [0, 0]
+
+
+def test_cat_dog_age_must_be_1_when_under_24() -> None:
+    assert get_human_age(24, 24) == [2, 2]
+
+
+def test_cat_dog_age_must_be_1_next_years() -> None:
+    assert get_human_age(28, 28) == [3, 2]
```
vsmutok, line 13: Please read [checklist.md](https://github.com/mate-academy/py-cat-and-dog-years/blob/master/checklist.md)

vsmutok (changes requested):

app/test_main.py
```diff
@@ -1,3 +1,23 @@
...
+        (0, 45, [0, 6]),
+        (48, 50, [8, 7]),
+    ]
+)
+def test_get_human_age(
+        cat_age: int,
+        dog_age: int,
+        expected_human_ages: list
+) -> None:
+    assert get_human_age(cat_age, dog_age) == expected_human_ages
```
vsmutok, line 23: Make sure you've tests checks all edged situations such as:

vsmutok (changes requested):

PR #941, opened 2024-05-08
triukhan solution

triukhan:


app/test_main.py
```diff
@@ -1,3 +1,25 @@
...
+    "cat_age,dog_age,converted_age_list", [
+        pytest.param(0, 0, [0, 0], id="zeros"),
+        pytest.param(14, 14, [0, 0], id="boundary less then one"),
+        pytest.param(15, 15, [1, 1], id="boundary more then one"),
+        pytest.param(23, 23, [1, 1], id="boundary less than two"),
+        pytest.param(24, 24, [2, 2], id="boundary more than two"),
+        pytest.param(27, 27, [2, 2], id="boundary less than three for cat"),
+        pytest.param(28, 28, [3, 2], id="boundary more than three for cat "
+                                        "and less than three for dog"),
+        pytest.param(100, 100, [21, 17], id="one hundred for each"),
```
LiudmylaKulinchenko, line 15: You probably could want to use larger numbers here to cover the case "if function resive data out of normal range, such as negative numbers, zero, or realy large numbers;"

LiudmylaKulinchenko (changes requested):
You've missed one point form the checklist: `if function resive data out of normal range, such as negative numbers, zero, or realy large numbers;`

app/test_main.py
triukhan: додав перевірку для realy large numbers

LiudmylaKulinchenko (changes requested):
Looks like we both missed this point from checklist: `Make sure you've tests checks all edged situations such as: if the function receives an incorrect type of data it raises the correct exception.` 👀 Add the test, please

LiudmylaKulinchenko (approved):
Nice job! 😸😸

PR #944, opened 2024-05-13
Animal testing

itpsychologist:


app/test_main.py
```diff
@@ -1,3 +1,13 @@
...
+def test_animal_not_first_year() -> None:
+    assert get_human_age(14, 12) == [0, 0]
+
+
+def test_animal_one_years() -> None:
+    assert get_human_age(15, 23) == [1, 1]
+
+
+def test_animal_two_years() -> None:
+    assert get_human_age(28, 28) == [3, 2]
```
vsmutok, line 13: * Make sure you've tests checks all edged situations such as:

* Use `@pytest.mark.parametrize`

vsmutok (changes requested):

EdAlekseiev (approved):

PR #945, opened 2024-05-14
Solution

VladimirDolhyi:


app/test_main.py
```diff
@@ -1,3 +1,33 @@
...
+def test_with_cat_more_24_and_dog_more_24() -> None:
+    assert get_human_age(25, 26) == [2, 2]
+
+
+def test_with_cat_28_and_dog_28() -> None:
+    assert get_human_age(28, 28) == [3, 2]
+
+
+def test_with_cat_50_and_dog_50() -> None:
+    assert get_human_age(50, 50) == [8, 7]
```
EdAlekseiev, line 33: Try to join these test functions to one with using parametrize. https://docs.pytest.org/en/7.1.x/how-to/parametrize.html

EdAlekseiev (changes requested):

app/test_main.py
VladimirDolhyi: Ok.

vasylhnatiuk (approved):

PR #947, opened 2024-05-18
implement py-cat-and-dog-years

olenazaritska:


vsmutok (changes requested):
image

checklist.md
```diff
@@ -1,4 +1,4 @@
-# Сheck Your Code Against the Following Points
+co# Сheck Your Code Against the Following Points
```
vsmutok, line 1: don't change checklist.md

vsmutok (changes requested):

checklist.md
olenazaritska: ops sorry that was accidental :)

app/test_main.py
```diff
@@ -1,3 +1,76 @@
...
+                   "if dog age is 24+"
+            ),
+            pytest.param(
+                100, 100, [21, 17],
+                id="add 1 human year per 4 cat years and 5 dog years "
+                   "if animal age is 24+"
+            ),
+            pytest.param(
+                -1, -1, [0, 0],
+                id="should return 0 if animal age is a negative integer"
```
Dimosphen1, line 47: Add test cases when you pass different values for `cat_age` and `dog_age` to check that the function processes correctly different inputs

app/test_main.py
```diff
@@ -1,3 +1,76 @@
...
+                id="should raise error "
+                   "if data type doesn't support arithmetic "
+                   "or comparison operators"
+            )
+        ]
+    )
+    def test_raise_error(
+            self,
+            cat_age: dict | set | str | list | tuple,
+            dog_age: dict | set | str | list | tuple,
```
Dimosphen1, line 73: There is no need to specify with such granularity the type annotation here, you could indicate that type could be `Any` instead

app/test_main.py
```diff
@@ -1,3 +1,76 @@
...
+        assert get_human_age(cat_age, dog_age) == expected_human_ages
+
+    @pytest.mark.parametrize(
+        "cat_age, dog_age",
+        [
+            pytest.param(
+                {"age": 3}, {"age": 12},
+                id="should raise error "
+                   "if data type doesn't support arithmetic "
+                   "or comparison operators"
```
Dimosphen1, line 66: Add test cases with tuples, lists, sets, and other data types, test not only dicts

Dimosphen1 (changes requested):

app/test_main.py
olenazaritska: Wouldn't that also include bool and float? Our function in main.py will not raise Error for such

PR #949, opened 2024-05-22
Hell yeah solution

RuslanLevchuk:


vasylhnatiuk (changes requested):
image
use Use @pytest.mark.parametrize, if the function receives an incorrect type of data it raises the correct exception.

app/test_main.py
```diff
@@ -1,3 +1,46 @@
...
+
+@pytest.mark.parametrize(
+    "cat_age,dog_age,expected",
+    [
+        (14, 12, [0, 0]),
+        (23, 23, [1, 1]),
+        (24, 24, [2, 2]),
+        (28, 29, [3, 3]),
+        (45, 57, [7, 8])
+    ]
```
vsmutok, line 43: what about negative or really large numbers?

vsmutok (changes requested):

app/test_main.py
RuslanLevchuk: > what about negative or really large numbers?
with negative numbers I agree. But how large numbers can be? There no specifiations in task.

vsmutok (approved):
Well Done!

PR #951, opened 2024-05-23
added tests for cat and dog years

AlexRozv:
solution


LiudmylaKulinchenko (approved):
Could be more tests for wrond types, but still looks great!

PR #953, opened 2024-05-24
Solution

rinarice:


EdAlekseiev (changes requested):
You should add tests for the next case
if function resive data out of normal range, such as negative numbers, zero, or realy large numbers

vsmutok (approved):
Well Done 👍

PR #955, opened 2024-05-27
Solution

SuskyiVolodymyr:


Oleg-s-s (changes requested):

DevSpaciX (changes requested):

app/test_main.py
```diff
@@ -1,3 +1,55 @@
...
+def test_convert_to_human_age(
+        cat_age: int,
+        dog_age: int,
+        human_age: list) -> None:
+    assert get_human_age(cat_age, dog_age) == human_age
+
+
+def test_convert_to_human_age_with_wrong_type() -> None:
+    with pytest.raises(TypeError):
+        get_human_age("20", "20")
```
Oleg-s-s, line 55:
```suggestion
pytest.param("20", "20", None, TypeError, id="Should raise TypeError if wrong type"),

def test_convert_to_human_age(cat_age: int, dog_age: int, human_age: list, exception: Exception) -> None:
    if exception:
        with pytest.raises(exception):
            get_human_age(cat_age, dog_age)
    else:
        assert get_human_age(cat_age, dog_age) == human_age
```

Oleg-s-s (approved):
This approach is ok, but you could simply add 1 more pytest param and check for exception

PR #958, opened 2024-05-29
Solution

TsvetanKichuk:


vasylhnatiuk (changes requested):

please read check list and fix your answer

app/test_main.py
```diff
@@ -1,3 +1,19 @@
...
+
+@pytest.mark.parametrize(
+    "cat_age, dog_age, result",
+    [
+        (0, 0, [0, 0]),  # test should return zero age
+        (14, 14, [0, 0]),  # test should return animals have not a year
+        (23, 23, [1, 1]),  # test should return animals have one year
+        (27, 27, [2, 2]),  # test should return animals have two years
+        (28, 28, [3, 2]),  # test should return animal have different age
+        (100, 100, [21, 17]),  # test should return animals have 100 years
```
Dimosphen1, line 14: Add test case when `cat_age` and `dog_age` have different values to test that output would not be affected

app/test_main.py
```diff
@@ -1,3 +1,19 @@
...
+        (-2, -5, [0, 0]),  # test should return if animal have negative age
+    ]
+)
+def test_modify(cat_age: int, dog_age: int, result: list) -> None:
+    assert get_human_age(cat_age, dog_age) == result
```
Dimosphen1, line 19: Add another function and use `pytest.mark.parametrize` to test the function in case strings, dicts, lists, sets, tuples and other data types are passed. You should also test against raised exceptions

app/test_main.py
```diff
@@ -1,3 +1,19 @@
+import pytest
+
 from app.main import get_human_age
 
-# write your code here
+
+@pytest.mark.parametrize(
+    "cat_age, dog_age, result",
+    [
+        (0, 0, [0, 0]),  # test should return zero age
```
Dimosphen1, line 9: There is no need to duplicate test content with comments, remove comments since data is self-explanatory

Dimosphen1 (changes requested):

app/test_main.py
```diff
@@ -1,3 +1,39 @@
...
+
+@pytest.mark.parametrize(
+    "cat_age, dog_age, result",
+    [
+        ("String", ("A", [1, 2, 3], False), TypeError),
+        ({dict}, {1, 2, 3}, TypeError)
+    ]
+)
+def test_incorrect_value(cat_age: int, dog_age: int, result: callable) -> None:
+    try:
```
pavlopro, line 31: Use context manager `pytest.raises` instead of this

pavlopro (changes requested):

PR #959, opened 2024-05-29
Solution

AndrewChugreev:


app/test_main.py
```diff
@@ -1,3 +1,49 @@
...
+            0,
+            29,
+            [0, 3],
+            id="add 1 for next 5 dog years"),
+        pytest.param(
+            -1,
+            -2,
+            [0, 0],
+            id="return 0 for negative value of age"),
+    ]
```
vsmutok, line 38: What about really large number?
Please add test for this case

vsmutok (changes requested):

app/test_main.py
```diff
@@ -1,3 +1,54 @@
...
+def test_convert_human_age(
+        cat_age: int,
+        dog_age: int,
+        human_age: int) -> None:
+    assert get_human_age(cat_age, dog_age) == human_age
+
+
+def test_convert_human_age_with_wrong_type_age() -> None:
+    with pytest.raises(TypeError):
+        get_human_age("20", "one")
```
Dimosphen1, line 54: Use `pytest.mark.parametrize` here as well to test not only against strings, but against dicts, lists, sets, tuples and other data types here

app/test_main.py
```diff
@@ -1,3 +1,54 @@
...
+            100,
+            100,
+            [21, 17],
+            id="return different values if cat_age = dog_age"),
+    ]
+)
+def test_convert_human_age(
+        cat_age: int,
+        dog_age: int,
+        human_age: int) -> None:
```
Dimosphen1, line 48: In case there are no arguments on the same line with an opening parenthesis, your closing parenthesis should not have any arguments on the same line as well. Please, resolve this issue here and in the `pytest.param(` as well

app/test_main.py
```diff
@@ -1,3 +1,54 @@
...
 
-# write your code here
+
+@pytest.mark.parametrize(
+    "cat_age,dog_age,human_age",
+    [
+        pytest.param(
+            5,
+            6,
+            [0, 0],
```
Dimosphen1, line 11: Placing arguments in several lines decreases the readability, place arguments in one line instead, please
```suggestion
            5, 6, [0, 0],
```

app/test_main.py
```diff
@@ -1,3 +1,54 @@
+import pytest
```
Dimosphen1, line 1: Add a blank line between different groups of import to follow PEP-8

Dimosphen1 (changes requested):

app/test_main.py
pavlopro: This was missed

pavlopro (changes requested):

pavlopro (approved):
Good Job!

PR #960, opened 2024-05-30
Solution

Shayba228:


EdAlekseiev (changes requested):
You should add tests for the next cases

app/test_main.py
```diff
@@ -1,3 +1,49 @@
...
+                         id="27/28 cat/dog years should "
+                         "convert into 2 human age."),
+            pytest.param(29, 28, [3, 3],
+                         id="28/29 cat/dog years should "
+                         "convert into 3 human age."),
+            pytest.param(-1, -1, [0, 0],
+                         id="-1 cat/dog years should "
+                            "convert into 0 human age."),
+        ]
+    )
```
vsmutok, line 31: add test of really large numbers

vsmutok (changes requested):

vsmutok (approved):
Great!

PR #963, opened 2024-06-03
'Solution'

BohdanKuzik:


Oleg-s-s (changes requested):
Add checks for negative values and wrong types

app/test_main.py
```diff
@@ -1,3 +1,28 @@
...
+        input1: int, input2: int, expected_result: list
+) -> None:
+    assert (
+        get_human_age(input1, input2) == expected_result
+    ), f"get_human_age({input1}, {input2}) should return {expected_result}"
+
+
+def test_get_human_age_exceptions() -> None:
+    with pytest.raises(TypeError):
+        get_human_age("1", "2")
```
DevSpaciX, line 28: Use here `@pytest.mark.parametrize` as you did in positive test case, but expect as a result value - some error for example . And read pls again comments of my colleague @wQuelS

DevSpaciX (changes requested):

DevSpaciX (approved):

PR #964, opened 2024-06-04
Solution

voboian:


app/test_main.py
```diff
@@ -1,3 +1,21 @@
+import pytest
 from app.main import get_human_age
 
-# write your code here
```
pavlopro, line 3: From the checklist:
If the function receives an incorrect type of data, it raises the correct exception.

Please cover this case as well

app/test_main.py
```diff
@@ -1,3 +1,21 @@
...
+
+@pytest.mark.parametrize(
+    "cat_age,dog_age,result",
+    [
+        (0, 0, [0, 0]),
+        (14, 14, [0, 0]),
```
pavlopro, line 9: Please use different values for dog and cat years

pavlopro (changes requested):

app/test_main.py
```diff
@@ -1,3 +1,40 @@
...
+        (1, "1", TypeError),
+        ("14", 13, TypeError),
+        (None, 15, TypeError),
+        (23, None, TypeError)
+    ]
+)
+def test_human_age_with_invalid_types(
+        cat_age: int,
+        dog_age: int,
+        error: type(Exception)
```
pavlopro, line 37: just Exception

pavlopro (approved):
Good Job!

PR #965, opened 2024-06-04
Solution

SrMusienko:


app/test_main.py
```diff
@@ -1,3 +1,19 @@
...
+        (15, 15, [1, 1]),
+        (23, 23, [1, 1]),
+        (24, 24, [2, 2]),
+        (27, 27, [2, 2]),
+        (28, 28, [3, 2]),
+        (100, 100, [21, 17])
+    ]
+)
+def test_get_human_age(cat_age: int, dog_age: int, result: list[int]) -> None:
+    assert get_human_age(cat_age, dog_age) == result
```
DevSpaciX, line 19: Add also negative tests cases : ) I mean where you expecting some error ot something

DevSpaciX (changes requested):

app/test_main.py
```diff
@@ -1,3 +1,25 @@
...
+        (-1, -1, [0, 0])
+    ]
+)
+def test_get_human_age(cat_age: int, dog_age: int, result: list[int]) -> None:
+    assert get_human_age(cat_age, dog_age) == result
+
+
+def test_incorrect_type_of_value() -> None:
+    with pytest.raises(TypeError):
+        get_human_age(2, "3")
```
DevSpaciX, line 25: Create some more cases with negative result . You can use `@pytest.mark.parametrize` here too

DevSpaciX (changes requested):

app/test_main.py
SrMusienko, line 1: The previous commit was stuck, I had to push it with a new commit

DevSpaciX (approved):

PR #970, opened 2024-06-05
Solution

necobra:


app/test_main.py
```diff
@@ -1,3 +1,41 @@
...
+
+
+def test_should_return_zeros_when_values_almost_15() -> None:
+    expected_value = [0, 0]
+    assert get_human_age(14, 14) == expected_value
+
+
+def test_should_return_ones_when_values_are_exactly_15() -> None:
+    expected_value = [1, 1]
+    assert get_human_age(15, 15) == expected_value
```
DevSpaciX, line 16: Instead of creating almost same tests you just can modify the parameters of test using `@pytest.mark.parametrize` . Aslo add negative tests cases , where you will expect some error for example : )

DevSpaciX (changes requested):

app/test_main.py
```diff
@@ -1,3 +1,35 @@
...
+        "human age equal two plus animal age divided by 4 for cat and by 5 for dog"
+    ]
+)
+def test_non_negative_inputs_get_human_age(cat_age: int, dog_age: int, cat_and_dog_age_in_human: list[int]) -> None:
+    assert get_human_age(cat_age, dog_age) == cat_and_dog_age_in_human
+
+
+def test_negative_inputs_get_human_age() -> None:
+    expected_value = [0, 0]
+    assert get_human_age(-3, -9) == expected_value
```
DevSpaciX, line 35: I mean you should add test where you will expect some error

app/test_main.py
```diff
@@ -1,3 +1,35 @@
...
+        "almost 1 human year",
+        "1 human year after 15 animal",
+        "almost 2 human year",
+        "2 human year after 24 animal",
+        "almost 3 human year",
+        "human age for cat is one more than for dog when animals ages are 28",
+        "human age equal two plus animal age divided by 4 for cat and by 5 for dog"
+    ]
+)
+def test_non_negative_inputs_get_human_age(cat_age: int, dog_age: int, cat_and_dog_age_in_human: list[int]) -> None:
```
DevSpaciX, line 29: This line is so long , pls split it on couple lines

DevSpaciX (changes requested):

app/test_main.py
necobra: the function we are testing cannot throw any error except the error when we pass data of the wrong type

DevSpaciX (approved):
Good job!

PR #973, opened 2024-06-06
cat-and-dog-years solution

DimaYankovskiy:


app/test_main.py
```diff
@@ -1,3 +1,48 @@
...
+            id="Really big numbers should work correctly."
+        )
+    ]
+)
+def test_cat_dog_age_to_human_age(
+        cat_age: int,
+        dog_age: int,
+        expected: list[int]
+) -> None:
+    assert get_human_age(cat_age, dog_age) == expected
```
DevSpaciX, line 48: Also add pls negative test case where you will expect some error

DevSpaciX (changes requested):

DevSpaciX (approved):
Nice!

PR #975, opened 2024-06-06
sol1

Unlie9:


app/test_main.py
```diff
@@ -1,3 +1,20 @@
...
+        (15, 15, [1, 1]),
+        (23, 23, [1, 1]),
+        (24, 24, [2, 2]),
+        (27, 27, [2, 2]),
+        (28, 28, [3, 2]),
+        (100, 100, [21, 17])
+    ]
+)
+def test_ages(cat_age: int, dog_age: int, expected: list[int, int]) -> None:
+    assert get_human_age(cat_age, dog_age) == expected
```
DevSpaciX, line 20: Add also negative test cases where you will expect some error for example

DevSpaciX (changes requested):

app/test_main.py
```diff
@@ -1,3 +1,23 @@
...
+        (24, 24, [2, 2], [1, 2]),
+        (27, 27, [2, 2], [3, 4]),
+        (28, 28, [3, 2], [3, 4]),
+        (100, 100, [21, 17], [23, 12])
+    ]
+)
+def test_expected_ages(cat_age: int, dog_age: int, expected: list[int, int],
+                       not_expected: list[int, int]) -> None:
+    assert get_human_age(cat_age, dog_age) == expected
+    assert get_human_age(cat_age, dog_age) != not_expected, ("Error, please "
```
DevSpaciX, line 22: Look , you did well , but for negative test cases it is better to create another test with another parametrized data ( there you can expect some errors or smth )

DevSpaciX (changes requested):

DevSpaciX (approved):

PR #977, opened 2024-06-06
Solution V.1

SemenetsSergii:


app/test_main.py
```diff
@@ -1,3 +1,48 @@
...
 from app.main import get_human_age
 
-# write your code here
+
+@pytest.mark.parametrize(
+    "cat_age, dog_age, result_age",
+    [
+        (-5, -5, [0, 0]),
+        (None, 5, TypeError),
+        ("1", 5, TypeError),
```
DevSpaciX, line 10: It is better to separate negative test cases into another test

DevSpaciX (changes requested):

DevSpaciX (approved):
Nice!

PR #988, opened 2024-06-07
Solution

Lebid-Dmytro:


pavlopro (changes requested):
From checklist:

if function recieve data such as negative

Polyakiv-Andrey (approved):
Good job!

PR #995, opened 2024-06-12
PC-cat-and-dog-years

DOLC3-GABBANA:


app/test_main.py
```diff
@@ -1,3 +1,33 @@
...
+        (15, -1, [1, 0]),
+        (-1, -1, [0, 0]),
+    ]
+)
+def test_get_human_age(
+    cat_age: int,
+    dog_age: int,
+    expected: get_human_age
+) -> None:
+    assert get_human_age(cat_age, dog_age) == expected
```
vsmutok, line 33: add test for this case:

vsmutok (changes requested):

EdAlekseiev (approved):

PR #999, opened 2024-06-17
solution

AndreiVed:


vasylhnatiuk (changes requested):
you should check if the function receives an incorrect type of data it raises the correct exception as well

vsmutok (approved):
Good Job!

PR #1009, opened 2024-06-26
+

AndriZhok:


app/test_main.py
```diff
@@ -1,3 +1,56 @@
 from app.main import get_human_age
+import pytest
 
-# write your code here
```
pavlopro, line 3: Make sure you've tests checks all edged situations such as:

pavlopro (changes requested):

app/test_main.py
AndriZhok: in the video on the subject of tests, it was clearly stated that writing tests to check the data type is the wrong approach, for this there are type annotations, the error in this case is the responsibility of the developer

Polyakiv-Andrey (approved):
Good job!

PR #1018, opened 2024-07-03
Solution

Muskam09:


NazarSheremet (changes requested):
image
Need to add test cases for this points

PR #1019, opened 2024-07-04
Solution

danilsiv:


vasylhnatiuk (changes requested):
you should check if the function receives an incorrect type of data it raises the correct exception as well

vasylhnatiuk (approved):

PR #1031, opened 2024-07-17
Solution

tylerj231:


app/test_main.py
```diff
@@ -1,3 +1,21 @@
 from app.main import get_human_age
-
+import pytest
 # write your code here
```
YukihiroSM, line 3: remove this comment, please.

YukihiroSM (changes requested):
Can you also try to perform test for parameters validity? 
For example, what if cat or dog years are string instead of int. 
Also, test that function works correctly with negative values.

vasylhnatiuk (approved):

vasylhnatiuk (approved):

PR #1037, opened 2024-07-22
implement tests_main

Morphin20th:


EdAlekseiev (changes requested):
Add tests for the next cases

app/test_main.py
```diff
@@ -1,3 +1,32 @@
...
+    ]
+)
+def test_can_convert_to_human_age(
+        cat_age: int,
+        dog_age: int,
+        result_list: list,
+        expected_error: None | Exception
+) -> None:
+    if expected_error:
+        with expected_error:
```
pavlopro, line 29: Please create a separate test method for expected errors

pavlopro (changes requested):
Good Job

PR #1039, opened 2024-07-23
Implemented test_main

OleksiukStepan:
Added tests:

- test_should_return_converted_amount_of_years
- test_should_work_with_large_numbers
- test_should_return_zero_to_negative_numbers
- test_with_incorrect_types
- test_with_not_enough_arguments

### DIFF
--- a/app/test_main.py
+++ b/app/test_main.py
@@ -59,14 +59,6 @@ def test_with_incorrect_types(cat_age: Any, dog_age: Any) -> None:
         get_human_age(cat_age, dog_age)
 
 
-@pytest.mark.parametrize(
-    "cat_age",
-    [
-        15,
-        ()
-    ],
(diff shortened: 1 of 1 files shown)


app/test_main.py
```diff
@@ -1,3 +1,72 @@
...
+
+@pytest.mark.parametrize(
+    "cat_age",
+    [
+        15,
+        ()
+    ],
+    ids=["1 argument", "0 arguments"]
+)
+def test_with_not_enough_arguments(cat_age: int) -> None:
```
oliasymak24, line 70: No need to use parametrize for only one case, u can write single test or add more cases

oliasymak24 (changes requested):

oliasymak24 (approved):

PR #1040, opened 2024-07-23
Implemented solution (separate check for cat and dog obliged me to complicate method with if)

dimak20:


app/test_main.py
```diff
@@ -1,3 +1,127 @@
...
+            assert get_human_age(
+                input_cat_data, input_dog_data
+            ) == expected_result
+            return
+        if not input_dog_data:
+            assert get_human_age(input_cat_data, 1)[0] == expected_result
+            return
+        if not input_cat_data:
+            assert get_human_age(1, input_dog_data)[1] == expected_result
+            return
```
MNDonut, line 88: You test cases when cat or dog years are None, but in annotation it's int. So can these params be None or you just test unreal cases?

MNDonut (changes requested):

app/test_main.py
dimak20: I want to check both variants separately. When someone sees that "test_dog" passes correctly while "test_cat" not, and vice versa, It'd be good for problem comprehesion and you do not need to change logic that passes test. Should I delete these two tests with None parametres?

oliasymak24 (approved):

PR #1041, opened 2024-07-25
Resolved Cat and Dog years

viktor-77:


app/test_main.py
```diff
@@ -1,3 +1,48 @@
+import pytest
+
 from app.main import get_human_age
```
pavlopro, line 3: Make sure you've tests checks all edged situations such as:

if function receive data out of normal range, such as negative numbers, zero, or realy large numbers;

PR #1044, opened 2024-07-29
'Solution'

PVS1905:


app/test_main.py
```diff
@@ -1,3 +1,13 @@
...
+def test_the_animal_is_one_year_old() -> None:
+    assert get_human_age(23, 15) == [1, 1]
+
+
+def test_the_animal_is_two_year_old() -> None:
+    assert get_human_age(24, 27) == [2, 2]
+
+
+def test_each_subsequent_year() -> None:
+    assert get_human_age(28, 29) == [3, 3]
```
vsmutok, line 13: Please read this file https://github.com/mate-academy/py-cat-and-dog-years/blob/master/checklist.md

vsmutok (changes requested):

EdAlekseiev (changes requested):
You should add tests for the next cases

vasylhnatiuk (approved):

PR #1045, opened 2024-07-30
Solution

lmidzhak:


vasylhnatiuk (changes requested):
add test for if the function receives an incorrect type of data it raises the correct exception.

PR #1046, opened 2024-07-30
Solution

DSahalatyi:


app/test_main.py
```diff
@@ -1,3 +1,59 @@
...
+    "animal_years, human_years",
+    [
+        ([24, 24], [2, 2]),
+        ([27, 28], [2, 2])
+    ]
+)
+def test_two_years_range_returns_two(
+        animal_years: list,
+        human_years: list
+) -> None:
```
MNDonut, line 44: you test the same function but with different params. you can implement 1 function that will test get_human_age but for each set of test data you can add ids parameter with a test data description

MNDonut (changes requested):

MNDonut (approved):

PR #1053, opened 2024-08-06
write tests for get_human_age function

vladislav-tsybuliak1:


app/main.py
```diff
@@ -1,4 +1,5 @@
 def get_human_age(cat_age: int, dog_age: int) -> list:
+    validate(cat_age, dog_age)
```
MNDonut, line 2: you are prohibited to change functions you are gonna test. the idea of tests is that somebody created some function and you test it regardless of it works correctly or not. your mission is just to test it with different sets of data

MNDonut (changes requested):

MNDonut (approved):

PR #1057, opened 2024-08-06
'Solution'

Losiev:


NazarSheremet (changes requested):
Good, but also need to add incorrect data type
image

NazarSheremet (approved):
Good job!

PR #1058, opened 2024-08-07
Solution

mihavryliuk:


app/test_main.py
```diff
@@ -1,3 +1,24 @@
...
+
+
 from app.main import get_human_age
 
-# write your code here
+
+class TestGetHumanAge:
+    @pytest.mark.parametrize(
+        "cat_age, dog_age, expected_result",
+        [
```
oliasymak24, line 10: Add tests for negative numbers and incorrect types

oliasymak24: Add one more test case if the function receives an incorrect type of data it raises the correct exception.

oliasymak24 (changes requested):

app/main.py
LiudmylaKulinchenko, line 1: No need to change main here

LiudmylaKulinchenko (changes requested):

app/main.py
mihavryliuk: if i dont change it it doesent pass the error test which i was talld to make

oliasymak24 (approved):

PR #1059, opened 2024-08-07
Task Solution

achalaidiuk:


app/test_main.py
```diff
@@ -1,3 +1,57 @@
...
+            "Twos for both at age under 28",
+            "Three for cat and two for dog at age 28",
+            "21 for cat and 17 for dog at age 100",
+        ]
+    )
+    def test_get_human_age(
+            self,
+            cat_age: int,
+            dog_age: int,
+            human_age: int
```
oliasymak24, line 36: Are you sure that  human_age wil be int?

oliasymak24 (changes requested):

PR #1060, opened 2024-08-07
Solution

Ivan-Shakhman:
Solution of task

### DIFF
--- a/app/test_main.py
+++ b/app/test_main.py
@@ -1,3 +1,43 @@
+import pytest
+
 from app.main import get_human_age
 
-# write your code here
+
+def test_should_return_list_with_two_values() -> None:
+    assert (
+        isinstance(get_human_age(1, 1), list)
+        and len(get_human_age(1, 1)) == 2
+    ), f"must return list but was {type(get_human_age(1, 1))}"
+
+
+def test_should_return_integers_numbers_in_list() -> None:
+    assert (
+        isinstance(num, int) for num in get_human_age(30, 30)
+    ), "human age must be integer"
+
+
+@pytest.mark.parametrize(
+    "cat_years,dog_years,expected_result",
+    [
+        (14, 14, [0, 0]),
+        (23, 23, [1, 1]),
+        (25, 27, [2, 2]),
+        (44, 51, [7, 7]),
+        (521, 1077, [126, 212])
+    ],
+    ids=[
+        "14 cats and 14 dogs years must convert to 0, 0 humans years",
+        "23 cats and 23 dogs years must convert to 1, 1 humans years",
+        "25 cats and 27 dogs years must convert to 2, 2 humans years",
+        "44 cats and 51 dogs years must convert to 7, 7 humans years",
+        "521 cats and 1077 dogs years must convert to 126, 212 humans years"
+    ]
+)
+def test_should_return_expected_result_for_each_param(
+    cat_years: int,
+    dog_years: int,
+    expected_result: list[int]
+) -> None:
+    assert (
+        get_human_age(cat_years, dog_years) == expected_result
+    ), f"must return {expected_result} when years is {cat_years, dog_years}"


app/test_main.py
```diff
@@ -1,3 +1,43 @@
...
+
+def test_should_return_list_with_two_values() -> None:
+    assert (
+        isinstance(get_human_age(1, 1), list)
+        and len(get_human_age(1, 1)) == 2
+    ), f"must return list but was {type(get_human_age(1, 1))}"
+
+
+def test_should_return_integers_numbers_in_list() -> None:
+    assert (
```
Oleksl888, line 14: your test data in below test checks that result is a list of integers - so you don't need this test really

app/test_main.py
```diff
@@ -1,3 +1,43 @@
+import pytest
+
 from app.main import get_human_age
 
-# write your code here
+
+def test_should_return_list_with_two_values() -> None:
+    assert (
```
Oleksl888, line 7: this test is also unnecessary

app/test_main.py
```diff
@@ -1,3 +1,43 @@
...
+def test_should_return_integers_numbers_in_list() -> None:
+    assert (
+        isinstance(num, int) for num in get_human_age(30, 30)
+    ), "human age must be integer"
+
+
+@pytest.mark.parametrize(
+    "cat_years,dog_years,expected_result",
+    [
+        (14, 14, [0, 0]),
```
Oleksl888, line 22: Recomendations from checlist are not fulfilled:

> 
> 
>     Make sure you've tests checks all edged situations such as:
> 
>     if output should change with some integer value you should check that it is not changed with previous value;
>     if function resive data out of normal range, such as negative numbers, zero, or realy large numbers;
>     if the function receives an incorrect type of data it raises the correct exception.
>

app/test_main.py
```diff
@@ -1,3 +1,43 @@
...
+    ]
+)
+def test_should_return_expected_result_for_each_param(
+    cat_years: int,
+    dog_years: int,
+    expected_result: list[int]
+) -> None:
+    assert (
+        get_human_age(cat_years, dog_years) == expected_result
+    ), f"must return {expected_result} when years is {cat_years, dog_years}"
```
Oleksl888, line 43: you don't need to add exception message here as you already set `ids` in parametrizer

app/test_main.py
```diff
@@ -1,3 +1,43 @@
...
+    ],
+    ids=[
+        "14 cats and 14 dogs years must convert to 0, 0 humans years",
+        "23 cats and 23 dogs years must convert to 1, 1 humans years",
+        "25 cats and 27 dogs years must convert to 2, 2 humans years",
+        "44 cats and 51 dogs years must convert to 7, 7 humans years",
+        "521 cats and 1077 dogs years must convert to 126, 212 humans years"
+    ]
+)
+def test_should_return_expected_result_for_each_param(
```
Oleksl888, line 36: Can you think of a better name for this test?

Oleksl888 (changes requested):

app/test_main.py
```diff
@@ -1,3 +1,34 @@
...
+        "14 cats and 14 dogs years must convert to 0, 0 humans years",
+        "23 cats and 23 dogs years must convert to 1, 1 humans years",
+        "25 cats and 27 dogs years must convert to 2, 2 humans years",
+        "44 cats and 51 dogs years must convert to 7, 7 humans years",
+        "521 cats and 1077 dogs years must convert to 126, 212 humans years"
+    ]
+)
+def test_should_correct_convert_animals_ages_to_humans(
+    cat_years: int,
+    dog_years: int,
```
LiudmylaKulinchenko, line 29: are you sure annotations should be int?

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Well done 🔥

PR #1063, opened 2024-08-07
commit

panichqa:


app/test_main.py
```diff
@@ -1,3 +1,58 @@
 from app.main import get_human_age
 
-# write your code here
+
```
oliasymak24, line 3: Use @pytest.parametrize

oliasymak24 (changes requested):

app/test_main.py
oliasymak24: Did you change the code regarding my comment?

oliasymak24 (changes requested):

app/test_main.py
```diff
@@ -1,3 +1,52 @@
...
+            (14, 0),
+            (15, 1),
+            (23, 1),
+            (24, 2),
+            (27, 2),
+            (28, 2),
+            (100, 17),
+        ]
+    )
+    def test_dog_age(self, dog_years: int, expected_human_years: int) -> None:
```
Oleksl888, line 51: No need to test cats and dogs separately :smile_cat: 
Also no need to create separate classes. Correct design for your tests would be:
1. One test class
2. Two methods - one for testing correct behaviour of the function and one that checks if correct exception has been raisedwhen data is incorrect.
3. Don't forget to include  `id/ids` for test data in parametrizer

Oleksl888 (changes requested):

app/test_main.py
```diff
@@ -1,3 +1,41 @@
...
+
+class TestGetHumanAge:
+    @pytest.mark.parametrize(
+        "cat_years, dog_years, expected",
+        [
+            (0, 0, [0, 0]),
+            (14, 14, [0, 0]),
+            (15, 0, [1, 0]),
+            (0, 15, [0, 1]),
+            (100, 100, [21, 17]),
```
oliasymak24, line 13: Add ids here

oliasymak24 (changes requested):
Your git tests were failed

oliasymak24 (changes requested):

LiudmylaKulinchenko (approved):
Good job 🤠

PR #1065, opened 2024-08-07
Solution of py-cat-and-dog-years

AndriiHamasa:


app/test_main.py
```diff
@@ -1,3 +1,36 @@
 from app.main import get_human_age
 
-# write your code here
+
```
oliasymak24, line 3: Use @pytest.parametrize

oliasymak24: Add tests for negative numbers and incorrect types

app/test_main.py
```diff
@@ -1,3 +1,40 @@
...
+        ),
+
+    ]
+)
+def test_get_human_age(
+        cat_age: int,
+        dog_age: int,
+        expected_age_list: list
+) -> None:
+    assert get_human_age(cat_age, dog_age) == expected_age_list
```
oliasymak24, line 40: Add one more test case if the function receives an incorrect type of data it raises the correct exception.

oliasymak24 (changes requested):

PR #1066, opened 2024-08-07
Solution

Bogdan-Rylov:


app/test_main.py
```diff
@@ -1,3 +1,56 @@
...
+        cat_age: int,
+        dog_age: int,
+        expected_human_age: list[int]
+) -> None:
+    assert get_human_age(cat_age, dog_age) == expected_human_age
+
+# In the checklist.md it is required to test "data out of normal range,
+# such as negative numbers, zero, or really large numbers"...
+
+# @pytest.mark.parametrize(
```
oliasymak24, line 35: Delete this commented out code

app/test_main.py
```diff
@@ -1,3 +1,56 @@
...
+)
+def test_converting_animal_age_to_human_age(
+        cat_age: int,
+        dog_age: int,
+        expected_human_age: list[int]
+) -> None:
+    assert get_human_age(cat_age, dog_age) == expected_human_age
+
+# In the checklist.md it is required to test "data out of normal range,
+# such as negative numbers, zero, or really large numbers"...
```
oliasymak24, line 33: Add tests for negative numbers and incorrect types

oliasymak24 (changes requested):

app/test_main.py
Bogdan-Rylov: I cannot push a solution with the tests for negative numbers and incorrect types because those tests don't pass. This is actually why I commented them

app/test_main.py
```diff
@@ -1,3 +1,30 @@
...
+
+@pytest.mark.parametrize(
+    "cat_age,dog_age,expected_human_age",
+    [
+        (14, 14, [0, 0]),
+        (15, 15, [1, 1]),
+        (23, 23, [1, 1]),
+        (24, 24, [2, 2]),
+        (27, 28, [2, 2]),
+        (28, 29, [3, 3]),
```
oliasymak24, line 14: You can add negative value directly here, not in separate function

oliasymak24 (changes requested):

oliasymak24 (approved):

PR #1068, opened 2024-08-08
first commit

egor250330:


app/test_main.py
```diff
@@ -1,3 +1,26 @@
+import pytest
+
 from app.main import get_human_age
 
-# write your code here
+
```
oliasymak24, line 5: Add tests for negative numbers and incorrect types

app/test_main.py
```diff
@@ -1,3 +1,26 @@
...
+        "dog_age, cat_age, excepted_age",
+        [
+            (0, 0, [0, 0]),
+            (14, 14, [0, 0]),
+            (15, 15, [1, 1]),
+            (23, 23, [1, 1]),
+            (24, 24, [2, 2]),
+            (27, 27, [2, 2]),
+            (28, 28, [3, 2]),
+            (100, 100, [21, 17]),
```
oliasymak24, line 17: Will be good to add some description for all cases

oliasymak24 (changes requested):

PR #1070, opened 2024-08-08
task done

Sergi0bbb:


app/test_main.py
```diff
@@ -1,3 +1,21 @@
...
+
+@pytest.mark.parametrize(
+    "cat_age, dog_age, expected_result",
+    [
+        (0, 0, [0, 0]),
+        (14, 14, [0, 0]),
+        (15, 15, [1, 1]),
+        (23, 23, [1, 1]),
+        (24, 24, [2, 2]),
+        (100, 100, [21, 17])
```
oliasymak24, line 13: Add tests for negative numbers and incorrect types

oliasymak24: Will be good to write some description for all cases

oliasymak24 (changes requested):

app/test_main.py
```diff
@@ -1,3 +1,33 @@
...
+        ),
+    ]
+)
+def test_get_human_age(
+        cat_age: int,
+        dog_age: int,
+        expected_result: int,
+        description: str
+) -> None:
+    assert get_human_age(cat_age, dog_age) == expected_result, description
```
oliasymak24, line 33: Add one more test case if the function receives an incorrect type of data it raises the correct exception.

oliasymak24 (changes requested):

app/test_main.py
```diff
@@ -1,3 +1,42 @@
...
+    ]
+)
+def test_get_human_age(
+        cat_age: int,
+        dog_age: int,
+        expected_result: int,
+        description: str,
+        expectation: pytest.fixture
+) -> None:
+    with expectation:
```
Oleksl888, line 41: It would be appropriate to have a separate test to check if correct exception is raised

Oleksl888 (changes requested):

app/test_main.py
```diff
@@ -1,3 +1,48 @@
...
+@pytest.mark.parametrize(
+    "cat_age, dog_age",
+    [
+        ("five", "five"),
+        (None, None),
+    ]
+)
+def test_human_age_errors(
+        cat_age: int,
+        dog_age: int,
```
LiudmylaKulinchenko, line 45: Looks like your annotations are wronk, since you want this test to recieve **any** incorrect type

app/test_main.py
```diff
@@ -1,3 +1,48 @@
...
+
+
+@pytest.mark.parametrize(
+    "cat_age, dog_age",
+    [
+        ("five", "five"),
+        (None, None),
+    ]
+)
+def test_human_age_errors(
```
LiudmylaKulinchenko, line 43: Let's try more descriptive name

LiudmylaKulinchenko (changes requested):

app/test_main.py
```diff
@@ -1,3 +1,48 @@
...
+            100,
+            [21, 17],
+            "When age 100 should return 21 for cat and 17 for dog",
+        )
+    ]
+)
+def test_get_human_age(
+        cat_age: int,
+        dog_age: int,
+        expected_result: int,
```
oliasymak24, line 30: You didn't check your annotations
Are you sure that expected_result is int?

oliasymak24 (changes requested):

app/test_main.py
```diff
@@ -1,3 +1,48 @@
...
+    [
+        (14, 14, [0, 0], "Age less than 15 should return [0, 0]"),
+        (23, 23, [1, 1], "Age equals to 23 return [1, 1]"),
+        (24, 24, [2, 2], "When age 24 should return [2, 2]"),
+        (-1, -1, [0, 0], "Negative ages should return [0, 0]"),
+        (
+            17,
+            17,
+            [1, 1],
+            "When age equals to 15 and less than 24 should return [1, 1]",
```
Oleksl888, line 17: you should use `id/ids` in parametrizer for text description - it is not a separate parameter and shouldnt be passed to test function explicitly

app/test_main.py
```diff
@@ -1,3 +1,48 @@
...
+) -> None:
+    assert get_human_age(cat_age, dog_age) == expected_result, description
+
+
+@pytest.mark.parametrize(
+    "cat_age, dog_age",
+    [
+        ("five", "five"),
+        (None, None),
+    ]
```
Oleksl888, line 41: `ids` here?

app/test_main.py
```diff
@@ -1,3 +1,48 @@
...
+)
+def test_human_age_for_invalid_input(
+        cat_age: str | None,
```
Oleksl888, line 44: `Any` would be correct type here

Oleksl888 (changes requested):

oliasymak24 (approved):

PR #1071, opened 2024-08-08
Resolve

ihorutkin:


app/test_main.py
```diff
@@ -1,3 +1,32 @@
...
+        [
+            (0, 0, 0, 0),
+            (14, 14, 0, 0),
+            (15, 15, 1, 1),
+            (23, 23, 1, 1),
+            (24, 24, 2, 2),
+            (27, 27, 2, 2),
+            (28, 28, 3, 2),
+            (100, 100, 21, 17),
+            (15.1, 15.5, 1, 1),
```
oliasymak24, line 19: Add tests for negative numbers and incorrect types

app/test_main.py
```diff
@@ -1,3 +1,32 @@
...
+        ]
+    )
+    def test_correct(
+            self,
+            cat_age: int,
+            dog_age: int,
+            expected_cat_human: int,
```
oliasymak24, line 26: expexted_years must be a list

app/test_main.py
oliasymak24: Add some description for all cases

oliasymak24 (changes requested):

app/test_main.py
```diff
@@ -1,3 +1,63 @@
...
+        assert get_human_age(cat_age, dog_age) == expected_years
+
+    @pytest.mark.parametrize(
+        "cat_age,dog_age,expected_error",
+        [
+            (3, None, TypeError),
+        ],
+        ids=[
+            "One of ages is None value",
+        ]
```
LiudmylaKulinchenko, line 54: Add a few more test cases. Otherwise you won't need a paramtrize

app/test_main.py
```diff
@@ -1,3 +1,63 @@
...
+    )
+    def test_incorrect(
+            self,
+            cat_age: int,
+            dog_age: int,
```
LiudmylaKulinchenko, line 59: Fix annotations, they are not ingers in this case

app/test_main.py
```diff
@@ -1,3 +1,63 @@
...
+            expected_error: BaseException
+    ) -> None:
+        with pytest.raises(expected_error):
+            get_human_age(cat_age, dog_age)
```
LiudmylaKulinchenko, line 63: If you test for the same type of error, no need to get it as a parameter, you can write it directly in the test

LiudmylaKulinchenko (changes requested):

oliasymak24 (approved):

PR #1072, opened 2024-08-08
Solution

mnezbrytska:


app/test_main.py
```diff
@@ -1,3 +1,33 @@
...
+    "cat_age, dog_age, expected",
+    [
+        (0, 0, [0, 0]),
+        (14, 14, [0, 0]),
+        (15, 15, [1, 1]),
+        (23, 23, [1, 1]),
+        (24, 24, [2, 2]),
+        (27, 27, [2, 2]),
+        (100, 100, [21, 17]),
+        (101, 101, [21, 17]),
```
oliasymak24, line 16: Add tests for negative numbers and also will be good to add some description for all cases

oliasymak24 (changes requested):

app/test_main.py
```diff
@@ -1,3 +1,48 @@
...
+        (-2, -3, [0, 0]),
+        (-2, 3, [0, 0]),
+        (14, 14, [0, 0]),
+        (15, 15, [1, 1]),
+        (23, 23, [1, 1]),
+        (24, 24, [2, 2]),
+        (100, 100, [21, 17])
+    ],
+    ids=[
+        "both values are zero",
```
Oleksl888, line 19: Use more informative descriptions here: e.g. `0 cat and dog age must equal 0 in human age`

app/test_main.py
```diff
@@ -1,3 +1,48 @@
...
+        "both values are negative",
+        "one value is negative",
+        "one animals year less than human year",
+        "first human year",
+        "23 not enough for two years",
+        "two human years",
+        "with big numbers"
+    ]
+)
+def test_get_human_age(cat_age: int, dog_age: int, expected: list) -> None:
```
Oleksl888, line 29: you can specify types of list contents as well

app/test_main.py
```diff
@@ -1,3 +1,48 @@
...
+
+@pytest.mark.parametrize(
+    "cat_age, dog_age",
+    [
+        ("two", -1),
+        ("ten", "five"),
+        (None, None),
+    ],
+    ids=[
+        "string and negative number",
```
Oleksl888, line 41: You can come up with better descriptions here too:
`TypeError raised when input is a string`

app/test_main.py
```diff
@@ -1,3 +1,48 @@
...
+        "two strings",
+        "two None type"
+    ],
+)
+def test_get_human_age_invalid_input(cat_age: int, dog_age: int) -> None:
+    with pytest.raises(TypeError):
```
Oleksl888, line 47: type annotations are wrong here since we expect not only ints

Oleksl888 (changes requested):

Oleksl888 (approved):
Perfect!

PR #1074, opened 2024-08-08
Solution

olena4reunited:


app/test_main.py
```diff
@@ -1,3 +1,60 @@
...
+                four for dog when inputs are twenty-eight"
+        ),
+        pytest.param(
+            100,
+            100,
+            [21, 17],
+            id="should return 21 for cat and \
+                17 for dog when inputs are one hundred"
+        ),
+    ]
```
oliasymak24, line 53: Add tests for negative numbers and incorrect types

oliasymak24 (changes requested):

app/test_main.py
```diff
@@ -1,3 +1,69 @@
...
+    assert get_human_age(cat_age, dog_age) == result, "Incorrect years convert"
+
+
+def test_raise_incorrect_type() -> None:
+    with pytest.raises(TypeError):
+        get_human_age("10", "10")
+
+
+def test_return_zeros_when_negative() -> None:
+    assert get_human_age(-10, -10) == [0, 0]
```
oliasymak24, line 69: You can add it to your first function with pytest.mark.parametrize

oliasymak24 (changes requested):

app/test_main.py
```diff
@@ -1,3 +1,71 @@
...
+            24,
+            24,
+            [2, 2],
+            id="should return twos when inputs are equal to twenty four"
+        ),
+        pytest.param(
+            28,
+            28,
+            [3, 2],
+            id="should return three for cat and \
```
Oleksl888, line 43: we don't use `\` as line breakers  as per style guidelines

app/test_main.py
```diff
@@ -1,3 +1,71 @@
...
+        ),
+        pytest.param(
+            -10,
+            -10,
+            [0, 0],
+            id="should return zeros when inputs less the then zero"
+        ),
+    ]
+)
+def test_correct_human_year_result(
```
Oleksl888, line 61: `test_get_human_age_works_correctly` would be a more suitable name

app/test_main.py
```diff
@@ -1,3 +1,71 @@
...
+    cat_age: int,
+    dog_age: int,
+    result: list[int]
+) -> None:
+    assert get_human_age(cat_age, dog_age) == result, "Incorrect years convert"
```
Oleksl888, line 66: no need to set error message since you already use `id` in parametrizer

app/test_main.py
```diff
@@ -1,3 +1,71 @@
...
+
+
+def test_raise_incorrect_type() -> None:
```
Oleksl888, line 69: `test_get_human_age_raises_exception` would be a more suitable name

app/test_main.py
```diff
@@ -1,3 +1,71 @@
...
+    with pytest.raises(TypeError):
+        get_human_age("10", "10")
```
Oleksl888, line 71: use parametrizer here as well

Oleksl888 (changes requested):

LiudmylaKulinchenko (approved):
Good job 🌷

PR #1075, opened 2024-08-09
Solution

NazarSerdiuk1:


app/test_main.py
```diff
@@ -1,3 +1,45 @@
...
+                        dog_human: int
+                        ) -> None:
+    assert get_human_age(cat_age, dog_age) == [cat_human, dog_human]
+
+
+@pytest.mark.parametrize(
+    "cat_age,dog_age",
+    [
+        pytest.param(-14, -14, id="negative age"),
+        pytest.param(100, 90, id="age exceeded"),
```
pavlopro, line 27: pytest.param(-14, -14, id="negative age"),
pytest.param(100, 90, id="age exceeded"),

Please those params to `test_the_animal_old` method params

Leave there only those which will produce errors

pavlopro (changes requested):

PR #1076, opened 2024-08-11
solving

Poznanskyi:


app/test_main.py
```diff
@@ -1,3 +1,57 @@
...
+            (14, 14, [0, 0]),
+            (15, 15, [1, 1]),
+            (23, 23, [1, 1]),
+            (24, 24, [2, 2]),
+            (27, 27, [2, 2]),
+            (28, 28, [3, 2]),
+            (100, 100, [21, 17]),
+        ],
+        ids=[
+            "Numbers should be positive",
```
Oleksl888, line 21: Use more informative descriptions here: e.g. `negative cat and dog age must equal 0 in human age`

app/test_main.py
```diff
@@ -1,3 +1,57 @@
...
+            (25, {}),
+        ],
+        ids=[
+            "TypeError for age list and string",
+            "TypeError for age dict",
+        ]
+    )
+    def test_incorrect_data(
+            self,
+            cat_age: int,
```
Oleksl888, line 53: it looks that here different types are expected

app/test_main.py
```diff
@@ -1,3 +1,57 @@
...
+
+    @pytest.mark.parametrize(
+        "cat_age, dog_age",
+        [
+            ([], "age"),
+            (25, {}),
+        ],
+        ids=[
+            "TypeError for age list and string",
+            "TypeError for age dict",
```
Oleksl888, line 48: `"TypeError raised when input is a dict"`

app/test_main.py
```diff
@@ -1,3 +1,57 @@
...
+        ]
+    )
+    def test_incorrect_data(
```
Oleksl888, line 51: `test_get_human_age_raises_exception` would more accurately describe what this test does

Oleksl888 (changes requested):

app/test_main.py
```diff
@@ -1,3 +1,57 @@
...
+        ],
+        ids=[
+            "TypeError for age list and string",
+            "TypeError raised when input is a dict",
+        ]
+    )
+    def test_get_human_age_raises_exception(
+            self,
+            cat_age: int | float,
+            dog_age: int | float
```
LiudmylaKulinchenko, line 54: In parametrize you don't pass float type, but you pass list, str, int and string. How can you add one annotation to cover them all? Let's find out!

LiudmylaKulinchenko (changes requested):

app/test_main.py
```diff
@@ -1,3 +1,59 @@
...
+        ],
+        ids=[
+            "TypeError for age list and string",
+            "TypeError raised when input is a dict",
+        ]
+    )
+    def test_get_human_age_raises_exception(
+            self,
+            cat_age: Union[int, float, list, str],
+            dog_age: Union[int, float, list, str]
```
LiudmylaKulinchenko, line 56: I meant you can use only _one_ type here, read about `Any` 👀

LiudmylaKulinchenko (changes requested):

PR #1084, opened 2024-08-18
py-cat-and-dog-years

s13en:
Solution?

### DIFF
--- a/app/test_main.py
+++ b/app/test_main.py
@@ -1,3 +1,29 @@
+import pytest
+
 from app.main import get_human_age
 
-# write your code here
+
+class TestCatDogAge:
+
+    @pytest.mark.parametrize("cat_age, dog_age, expected", [
+        (14, 14, [0, 0]),
+        (15, 15, [1, 1]),
+        (23, 23, [1, 1]),
+        (25, 25, [2, 2]),
+        (28, 29, [3, 3]),
+        (32, 34, [4, 4]),
+        (100, 100, [21, 17]),
+        (-1, -5, [0, 0]),
+    ],
+        ids=[
+            "14 cat/dog years should convert into 0 human age.",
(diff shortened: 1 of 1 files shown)


s13en:
I don't get it at all

app/test_main.py
```diff
@@ -1,3 +1,47 @@
...
+            dog_age: int,
+            expected: list
+    ) -> None:
+        assert get_human_age(cat_age, dog_age) == expected
+
+    @pytest.mark.parametrize(
+        "cat_age, dog_age",
+        [
+            (-1, -5),
+            (150, 200),
```
Dimosphen1, line 38: Negative numbers, 0, large numbers and numbers different for cat and dog age are all valid values and they should be included in the `test_get_human_age`. They do not raise `TypeError`. Test `get_human_age` against strings, dicts, tuples and other data types, they should raise `TypeError` during `get_human_age` execution.

Dimosphen1 (changes requested):

pavlopro (changes requested):

app/test_main.py
```diff
@@ -1,3 +1,42 @@
...
+    )
+    def test_get_human_age(
+            self,
+            cat_age: int,
+            dog_age: int,
+            expected: list
+    ) -> None:
+        assert get_human_age(cat_age, dog_age) == expected
+
+    def test_get_human_age_typeerror(self) -> None:
```
LLkaia, line 40: Let's test not only with string but with dict, tuple, list as well. Use parametrize decorator here

LLkaia (changes requested):

LLkaia (approved):
Great job!

Pay attention: `{5}` is not a **dict**, it is **set**.

PR #1088, opened 2024-08-22
implement py-cat-and-dog-years

dkn3r:


vasylhnatiuk (changes requested):
Use @pytest.mark.parametrize

app/test_main.py
```diff
@@ -1,3 +1,22 @@
...
+        (14, 14, [0, 0]),
+        (15, 15, [1, 1]),
+        (23, 23, [1, 1]),
+        (24, 24, [2, 2]),
+        (27, 27, [2, 2]),
+        (28, 28, [3, 2]),
+        (100, 100, [21, 17])
+    ]
+)
+def test_get_human_age(cat_age: int, dog_age: int, expected: list) -> None:
```
vasylhnatiuk, line 18:
```suggestion
def test_get_human_age(cat_age: int, dog_age: int, expected: list[int]) -> None:
```

vasylhnatiuk (changes requested):

PR #1096, opened 2024-08-28
Solution

juliazelena:


app/test_main.py
```diff
@@ -1,3 +1,52 @@
 from app.main import get_human_age
```
pavlopro, line 2: Please use `pytest.mark.parametrize` decorator as mentioned in task description

pavlopro (changes requested):

LLkaia (changes requested):
Add test with incorrect data, e.g. which will though TypeError

PR #1106, opened 2024-09-04
cat and dog years

AndreyHMkr:


app/test_main.py
```diff
@@ -1,3 +1,82 @@
+import pytest
 from app.main import get_human_age
 
-# write your code here
+
+class TestResultAnimalAge:
+    @pytest.mark.parametrize(
+        "cat_age,dog_age,result",
+        [
+            (
```
pavlopro, line 9: Please format code to be more readable. 
```
(0, 0, [0, 0]),
(14, 14, [0, 0]),
```
and so on..

pavlopro (changes requested):

app/test_main.py
```diff
@@ -1,3 +1,124 @@
...
+                0,
+                0,
+                [0, 0]
+            ),
+            (
+                14,
+                14,
+                [0, 0]
+            ),
```
LLkaia, line 18:
```suggestion
            (14, 14, [0, 0]),
```
And same for all cases bellow

LLkaia (changes requested):

LLkaia (approved):
GJ!

PR #1110, opened 2024-09-10
Solution

CheshireKate:


app/test_main.py
```diff
@@ -1,3 +1,48 @@
+import pytest
```
Dimosphen1, line 1: Add one blank line between different import groups to follow PEP-8

app/test_main.py
```diff
@@ -1,3 +1,48 @@
...
+            200, 200, [46, 37],
+            id="big numbers"
+        )
+    ]
+)
+def test_get_human_age(cat_age: int, dog_age: int, result: list) -> None:
+    assert isinstance(cat_age, int) and isinstance(dog_age, int),\
+        "Age should be an integer."
+    assert cat_age >= 0 and dog_age >= 0,\
+        "Age should be a positive number or 0."
```
Dimosphen1, line 45: There is no need for these assert statements, the `get_human_age` should work with negative values as well. You should add a separate function that tests against raised errors in case data types other than integers are passed, e.g. lists, strings, tuples, dicts, etc.

app/test_main.py
```diff
@@ -1,3 +1,48 @@
...
+            id="cat's third year border"
+        ),
+        pytest.param(
+            28, 28, [3, 2],
+            id="difference between animals"
+        ),
+        pytest.param(
+            200, 200, [46, 37],
+            id="big numbers"
+        )
```
Dimosphen1, line 38: Add test cases with negative numbers and different values for `cat_age` and `dog_age` to make sure that final values are correct in such cases as well

Dimosphen1 (changes requested):

PR #1112, opened 2024-09-11
solution

uzlss:


YukihiroSM (changes requested):
Your test is good! 

Could you also test the function behaviour when the values are incorrect? 

For example, numbers are negative (-1) of numbers are strings instead of integers etc.

uzlss:
> Your test is good!
> 
> Could you also test the function behaviour when the values are incorrect?
> 
> For example, numbers are negative (-1) of numbers are strings instead of integers etc.

main.py does not raise exception for negative numbers :P

PR #1127, opened 2024-09-25
done

ivanbilobrov07:


NazarSheremet (changes requested):
image
Need to add tests for incorrect data

NazarSheremet (approved):
Good job!

PR #1130, opened 2024-09-30
Solution

anastasiiakalini:


app/test_main.py
```diff
@@ -1,3 +1,24 @@
...
+            (28, 28, [3, 2]),
+            (100, 100, [21, 17])
+        ]
+    )
+    def test_check_age_for_cat_and_dog_in_human_age(self,
+                                                    age_cat: int,
+                                                    age_dog: int,
+                                                    expected_result: list
+                                                    ) -> None:
+        assert get_human_age(age_cat, age_dog) == expected_result
```
pavlopro, line 24: Please take into consideration points from checklist:

app/test_main.py
```diff
@@ -1,3 +1,24 @@
...
+            (14, 14, [0, 0]),
+            (15, 15, [1, 1]),
+            (23, 23, [1, 1]),
+            (24, 24, [2, 2]),
+            (27, 27, [2, 2]),
+            (28, 28, [3, 2]),
+            (100, 100, [21, 17])
+        ]
+    )
+    def test_check_age_for_cat_and_dog_in_human_age(self,
```
pavlopro, line 19: Please format code:

```
def test_check_age_for_cat_and_dog_in_human_age(
    self,
    age_cat: int,
    ...
)
```

pavlopro (changes requested):

app/test_main.py
anastasiiakalini: What is wrong with my new tests?

pavlopro (approved):
GJ!

PR #1133, opened 2024-10-03
Cat and Dog years

vladyslavgusakovskiy:


NazarSheremet (changes requested):
image
Add test case for incorect input data

NazarSheremet (approved):
Good job!

PR #1142, opened 2024-10-10
Solution

VladyslavBon:


app/test_main.py
```diff
@@ -1,3 +1,61 @@
...
+        ),
+        pytest.param(
+            28, 29, [3, 3],
+            id="every 5 next dog years should give 1 extra human year"
+        ),
+        pytest.param(
+            0, 0, [0, 0],
+            id="should return zeros when cat or dog age equals zeros"
+        ),
+    ]
```
Nattalli, line 29: add more tests:
In ideal scenario, you should test for all the boundaries between different outputs, for example, for cat years:

0;
14: largest possible value that returns 0 human years;
15: smallest possible value that returns 1 human year;
23: largest possible value that returns 1 human year; and so on.
This does not apply to numbers past 28, since they increase linearly, and you can just test for a large value.

app/test_main.py
```diff
@@ -1,3 +1,61 @@
...
+def test_handle_correct_data_input(
+        cat_age: int,
+        dog_age: int,
+        expected_error: Exception
+) -> None:
+    with pytest.raises(expected_error):
+        assert isinstance(cat_age, int), "cat_age must be integer"
+        assert isinstance(dog_age, int), "cat_age must be integer"
+        assert 0 <= cat_age <= 100, "cat_age must be in range from 0 to 100"
+        assert 0 <= dog_age <= 100, "dog_age must be in range from 0 to 100"
```
Nattalli, line 61: better call function and assert function errors:
```python
with pytest.raises(TypeError):
        get_human_age(...)
```

Nattalli (changes requested):

app/test_main.py
```diff
@@ -1,3 +1,102 @@
...
+    ]
+)
+def test_handle_correct_data_input_value(
+        cat_age: int,
+        dog_age: int,
+        expected_error: AssertionError
+) -> None:
+    with pytest.raises(AssertionError):
+        assert 0 <= cat_age <= 100, "cat_age must be in range from 0 to 100"
+        assert 0 <= dog_age <= 100, "dog_age must be in range from 0 to 100"
```
AnyoneClown, line 102: We can refactor this using only 1 testcase. Create 1 test and name it smth like `test_handle_data_input`.

We this we can handle both errors(`TypeError` and `AssertionError`) in 1 test. You should do this because you have similar logic for this tests, both of them are handling input and have the same structure in `pytest.mark.parametrize`

AnyoneClown (changes requested):

Arsen-hrynevych (approved):
ok

PR #1144, opened 2024-10-13
cat_and_dog_years solution

94nj111:


app/test_main.py
AnyoneClown, line 1: Add one more test caseif the function receives an incorrect type of data it raises the correct exception.

app/test_main.py
```diff
@@ -1,3 +1,22 @@
+import pytest
+
 from app.main import get_human_age
 
-# write your code here
+
+@pytest.mark.parametrize(
+    "cat_age,dog_age,result",
+    [
+        (0, 0, [0, 0]),
```
AnyoneClown, line 9: Add tests for negative numbers and incorrect types

app/test_main.py
```diff
@@ -1,3 +1,22 @@
...
+        (14, 14, [0, 0]),
+        (15, 23, [1, 1]),
+        (27, 27, [2, 2]),
+        (28, 28, [3, 2]),
+        (100, 100, [21, 17])
```
AnyoneClown, line 14: Add comments for all test cases, so user can understand why tests didn't pass.

You can use argument `id` for each of your tests. Read about this:
[Link](https://docs.pytest.org/en/stable/example/parametrize.html#set-marks-or-test-id-for-individual-parametrized-test)

AnyoneClown (changes requested):

AnyoneClown (approved):
LGTM!

PR #1151, opened 2024-10-17
Solution

molodsh1y1:


yulia-pl (approved):

app/main.py
```diff
@@ -1,4 +1,8 @@
 def get_human_age(cat_age: int, dog_age: int) -> list:
```
Arsen-hrynevych, line 1: The function could be more concise. Check the return statement.

app/main.py
```diff
@@ -12,3 +16,7 @@ def convert_to_human(
     if animal_age < first_year + second_year:
```
Arsen-hrynevych, line 16: Change the second if to elif which is typically more efficient and clearer.

Arsen-hrynevych (approved):
Ok

PR #1154, opened 2024-10-17
solution

serhiik05:


pohgen (approved):
good, but you can also add check for border age cat and dog
f.e. [2, 3]

PR #1172, opened 2024-10-20
solution

YaYaYaroslav:


MatvijSafonov (approved):
Better to do it with parametrize. But anyway looking good

PR #1174, opened 2024-10-21
Solution

Icemann007:


vasylhnatiuk (changes requested):
if the function receives an incorrect type of data it raises the correct exception

app/main.py
```diff
@@ -1,4 +1,10 @@
 def get_human_age(cat_age: int, dog_age: int) -> list:
+    if not isinstance(cat_age, int) or not isinstance(dog_age, int):
```
vasylhnatiuk, line 2: don't change main code and it is useless

vasylhnatiuk (changes requested):

app/main.py
Icemann007: How should I implement the function so that if it receives the wrong data type, it throws the correct exception?

vasylhnatiuk (approved):

PR #1253, opened 2024-12-18
solution

DanilTrukhanov:


4ikkamony (approved):
Good job!

The task's checklist mentioned, that tests must also check if correct exception is raised if function receives incorrect datatypes.
But as the function doesn't do it explicitly, I'm not sure if test for it should be included. Consider it)

DanilTrukhanov:
> Good job!
> 
> The task's checklist mentioned, that tests must also check if correct exception is raised if function receives incorrect datatypes. But as the function doesn't do it explicitly, I'm not sure if test for it should be included. Consider it)

@4ikkamony Yeah, I was thinking about your correction, but I remembered the lecture where the teacher said if a function has type annotations - passing arguments with the wrong type is a problem of who did that) 
maybe I'm wrong))

4ikkamony:
@DanilTrukhanov fair point)

PR #1254, opened 2024-12-18
Solution

4ikkamony:


bogdAAAn1 (approved):
A cool and unusual solution!
Good job!

PR #1274, opened 2024-12-20
solution

MaksVakulenko:


L1nk3rrr (changes requested):
`Please note: you have to use pytest for writing tests.`
Think how you can implement and use pytest here(parameterize for example, also add message for your asserts)

L1nk3rrr (changes requested):
don't forget about assert messages, also such styling 
```
@pytest.mark.parametrize(
     "cat_age, dog_age, expected", 
     [
        (0, 0, [0, 0]),
        (14, 14, [0, 0]),
      ]
)
```

L1nk3rrr (approved):
Next time, do not forget about assert messages.
Well done!

PR #1275, opened 2024-12-21
Created all tests

shafran-andriy:


vsmutok (changes requested):
1. Make sure you've tests checks all edged situations such as:
* if output should change with some integer value you should 
check that it is not changed with previous value;
* if function resive data out of normal range, 
such as negative numbers, zero, or realy large numbers;

2. Use ```@pytest.mark.parametrize```

vsmutok (changes requested):
Add tests for this case:
> if the function receives an incorrect type of data it raises the correct exception.

vsmutok (approved):
Great!

PR #1338, opened 2025-02-05
complete tests for human_age func. 

kram3ko:
Using class, pytest.mark.parametrize. and for example used expected error


Baranotik15 (approved):
good job!

PR #1374, opened 2025-02-26
Finish

Sotnikov-100:


yarzhuravlov (changes requested):
Good job! But in this task, using `@pytest.mark.parametrize` is mandatory. 

Last item on the checklist

PR #1384, opened 2025-03-02
Completed task cat-and-dog-years

Yurkiv313:


app/test_main.py
oliasymak24, line 1: Add tests for negative numbers and incorrect types

oliasymak24 (changes requested):

Boyok1337 (approved):
Good job! Try to correctly name the commits that have been made/added/deleted, not just fix_1, fix_2

PR #1388, opened 2025-03-03
Solution

Oneyura:


app/test_main.py
```diff
@@ -1,3 +1,37 @@
...
+        (-1, -1, [0, 0]),
+        (15, 28, [1, 2]),
+
+    ],
+    ids=[
+        "14 years for cat and dog",
+        "15 years for cat and dog",
+        "24 years for cat and dog",
+        "28 years for cat and dog",
+        "1000 years for cat and dog",
```
AndyStarGitHub, line 25: There was a request in the checklist: if function resive data out of normal range, such as negative numbers, zero, or realy large numbers; perhaps we should not accept 1000 years?

AndyStarGitHub (approved):
Just one comment (1000 years) for the future.

PR #1389, opened 2025-03-04
'Solution'

AndyStarGitHub:


app/test_main.py
```diff
@@ -1,3 +1,44 @@
...
+
+@pytest.mark.parametrize(
+    "cat_age,dog_age",
+    [
+        (1, 0),
+        (14, 2),
+        (111, 15),
+        (24, 120)]
+)
+def test_age_is_realistic(cat_age: int, dog_age: int) -> None:
```
Oneyura, line 35: This test will never raise exception, think about that.
It looks like program dont have that option, what you want to test.

app/test_main.py
```diff
@@ -1,3 +1,44 @@
...
+    "cat_age,dog_age,human_age",
+    [
+        (0, 0, [0, 0]),
+        (14, 14, [0, 0]),
+        (15, 15, [1, 1]),
+        (24, 24, [2, 2]),
+        (27, 27, [2, 2]),
+        (28, 28, [3, 2]),
+        (100, 100, [21, 17])
+    ]
```
Oneyura, line 16: Here after value for test you can add ids =[ ] for more information about each test.

Oneyura (approved):
I am not sure about some test that you have made, but if AI approve this, then everything is good.

PR #1390, opened 2025-03-04
Implement tests for get_human_age function

PrimeGlorious:


app/test_main.py
```diff
@@ -1,3 +1,46 @@
...
-# write your code here
+
+@pytest.mark.parametrize(
+    "cat_years, dog_years, human_age_list",
+    [
+        pytest.param(
+            0,
+            0,
+            [0, 0],
+            id="test should work correctly with zeros"
```
BronkstoneBro, line 13: I'm not quite sure why you're moving each parameter to a separate line; it seems harder to read. 
The ID can be moved out of each parameter into ids=. 

Everything else is fine.

BronkstoneBro (approved):

Boyok1337 (approved):

PR #1406, opened 2025-03-07
'Solution'

pewlay:


app/test_main.py
```diff
@@ -1,3 +1,13 @@
...
+def test_should_return_0_when_age_less_than_15() -> None:
+    assert get_human_age(15, 14) == [1, 0]
+
+
+def test_should_return_1_when_age_less_than_24() -> None:
+    assert get_human_age(24, 23) == [2, 1]
+
+
+def test_should_return_array_when_age_bigger_29() -> None:
+    assert get_human_age(27, 29) == [2, 3]
```
Boyok1337, line 13: Not enought test, try to add more tests, also pay attention to `@pytest.mark.parametrize` approach with ids to recognize error in tests. 
Also, good to add negative values

Boyok1337 (changes requested):

Boyok1337 (approved):
GJ!

PR #1414, opened 2025-03-09
Solution cat-and-dog-years

Roksolana-K:


app/test_main.py
oliasymak24, line 1: Add one more test case if the function receives an incorrect type of data it raises the correct exception.

oliasymak24 (changes requested):

oliasymak24 (approved):

PR #1451, opened 2025-03-31
Solution

Tolik-code:


farzamgt (approved):
I hadn't thought of that approach before—very interesting!

app/test_main.py
```diff
@@ -1,3 +1,162 @@
...
+    )
+    def test_get_human_age(
+            self,
+            cat_age: int,
+            dog_age: int,
+            expected: [int]
+    ) -> None:
+        assert (
+            get_human_age(cat_age, dog_age) == expected
+        )
```
L1nk3rrr, line 53: looks like it can be one lined
`assert get_human_age(cat_age, dog_age) == expected`

L1nk3rrr (approved):
Amazing work was done!

PR #1457, opened 2025-04-03
Solution

farzamgt:


app/test_main.py
```diff
@@ -1,3 +1,62 @@
 from app.main import get_human_age
 
-# write your code here
+
+def test_get_human_age_when_both_ages_are_zero() -> None:
+    result = get_human_age(0, 0)
+    assert result == [0, 0]
```
Tolik-code, line 6: Better not create extra variables, it took additional memory, can use "assert get_human_age(0, 0) == [0, 0]"

PR #1587, opened 2025-07-11
solution py cat and dog years

PavelSemenikhin:


app/test_main.py
```diff
@@ -1,3 +1,22 @@
...
+        (14, 14, [0, 0]),
+        (23, 23, [1, 1]),
+        (24, 24, [2, 2]),
+        (27, 27, [2, 2]),
+        (28, 28, [3, 2]),
+
+    ]
+)
+def test_cat_dog_years_to_human(cat_age: int,
+                                dog_age: int, expected: list[int]) -> None:
```
L1nk3rrr, line 20: Here can be better functioning styling, like this:
```
def test_cat_dog_years_to_human(
    cat_age: int, 
    dog_age: int, 
    expected: list[int]
) -> None:
```

Use a one liner for arguments if it is possible, if not -> one argument per line

L1nk3rrr (approved):
Well done! 

But remember rule about styling!

PR #1588, opened 2025-07-11
Solution

Kostenikov:


SensibleN00B (approved):
Clean Code
Add comments, prints, and functions to check your solution when you write your code. Don't forget to delete them when you are ready to commit and push your code.

Use @pytest.mark.parametrize

PR #1604, opened 2025-07-16
Solution

bohdankolisnyk04:


app/test_main.py
```diff
@@ -1,3 +1,45 @@
...
+
+
+def test_cat_age_should_return_correct_value_for_large_age() -> None:
+    assert get_human_age(100, 0)[0] == 21
+
+
+def test_dog_age_should_return_0_when_less_than_15() -> None:
+    assert get_human_age(0, 14)[1] == 0
+
+
```
MNDonut, line 27: Let's use `parametrize` because all your tests have the same body, the only difference is parameters and an expected result

MNDonut (changes requested):

PR #1805, opened 2025-12-19
Solution

herasymchukkdenysss:


herasymchukkdenysss:
there is no need to check whether the function handle incorrect input. The function uses type annotation, and it's bound to function user whether give a correct defined params or not(an IDE should hint him). (It was said in the video from mateacademy)

PR #1853, opened 2026-01-23
Add tests for cat and dog human age calculation

nazarkravets2007:


nazarkravets2007 (commented):
Refactor tests using parametrize and add edge case coverage

PR #1877, opened 2026-02-08
all tests passed

Mykola-Osolinskyi:


app/main.py
```diff
@@ -1,23 +1,21 @@
...
-    Examples:
-        get_human_age(0, 0) == [0, 0]
-        get_human_age(15, 15) == [1, 1]
-        get_human_age(24, 24) == [2, 2]
-    """
-    # TODO: Implement this function
-    # Write your tests first, then implement the logic
-    return [0, 0]
+def get_human_age(cat_age: int, dog_age: int) -> list[int]:
+    def convert_to_human(animal_years: int, step: int) -> int:
```
oliasymak24, line 2: You need to check type of incoming parameters as you have a test function for this

oliasymak24, line 2: Why you need inner function?

oliasymak24 (changes requested):

app/main.py
Mykola-Osolinskyi: add required check

Mykola-Osolinskyi: it looks more structured for me)

oliasymak24 (approved):

PR #1912, opened 2026-02-25
cat and dog years

KostiantinPastukh:


app/main.py
```diff
@@ -1,23 +1,46 @@
...
+        elif i == 1 and dog_age >= 9:
+            result[1] += 1
+            dog_age -= 9
+        elif i > 1:
+            result[1] += dog_age // 5
+            break
+        else:
+            break
+
+    return result
```
vsmutok, line 46: Try to avoid using loops here 🙂

vsmutok (changes requested):

PR #1915, opened 2026-03-02
Solution: written tests and then function

SerhiiHryniuk:


app/main.py
```diff
@@ -1,23 +1,34 @@
...
     Returns:
         List with [cat_human_age, dog_human_age]
-        
     Examples:
         get_human_age(0, 0) == [0, 0]
         get_human_age(15, 15) == [1, 1]
         get_human_age(24, 24) == [2, 2]
     """
-    # TODO: Implement this function
-    # Write your tests first, then implement the logic
```
Reverad, line 22: that looks terrible

app/main.py
```diff
@@ -1,23 +1,34 @@
...
+        cat_human_age = 0
+    elif cat_age < 24:
+        cat_human_age = 1
+    else:
+        cat_human_age = 2 + (cat_age - 24) // 4
+
+    if dog_age < 15:
+        dog_human_age = 0
+    elif dog_age < 24:
+        dog_human_age = 1
```
Reverad: here's duplication

app/test_main.py
```diff
@@ -1,3 +1,34 @@
...
+
+@pytest.mark.parametrize(
+    "cat_age, dog_age, expected_age",
+    [
+        (-1, -1, [0, 0]),
+        (0, 0, [0, 0]),
+        (14, 14, [0, 0]),
+        (23, 23, [1, 1]),
+        (28, 28, [3, 2]),
+        (100, 100, [21, 17])
```
Reverad, line 19: u could add few scenarios with different numbers

Reverad (changes requested):
overall it looks good(especially tests), but u need to fix a few things

app/main.py
SerhiiHryniuk: What exactly there is terrible? I did not write that it was before me

PR #1967, opened 2026-04-02
Implemented solution for py-cat-and-dog-years and test for this task

Neyrony:


Neyrony:
I think func to check data is unneccesary because i have func anotation and that is not my fault when someone gives falty vars. That was mention in video

PR #1972, opened 2026-04-03
solved

pawunder1992:


app/test_main.py
```diff
@@ -1,3 +1,47 @@
+import pytest
 from app.main import get_human_age
```
VadimPusiak788, line 3: You should separate imports by group and separate them with blank lines

1. Standard library imports
2. Related third-party imports
3. Local application/library imports

app/main.py
```diff
@@ -1,23 +1,30 @@
...
-        get_human_age(0, 0) == [0, 0]
-        get_human_age(15, 15) == [1, 1]
-        get_human_age(24, 24) == [2, 2]
-    """
-    # TODO: Implement this function
-    # Write your tests first, then implement the logic
-    return [0, 0]
+    if (not isinstance(cat_age, (int, float))
+            or not isinstance(dog_age, (int, float))):
+        raise TypeError
```
VadimPusiak788: Please  always raise exceptions with a descriptive message:

app/main.py
```diff
@@ -1,23 +1,30 @@
...
+    cat_age = int(cat_age)
+    dog_age = int(dog_age)
+    list_of_years = [15, 9, 4]
+    result = [0, 0]
+    for i in list_of_years:
```
VadimPusiak788, line 15: There is duplication between these two loops. It would be cleaner and easier to maintain if this logic were extracted into a helper function parameterized by the age and step values.

VadimPusiak788 (changes requested):

app/test_main.py
pawunder1992: ok, done

app/main.py
pawunder1992: fixed

app/main.py
pawunder1992: done

app/main.py
```diff
@@ -1,30 +1,25 @@
 def get_human_age(cat_age: int, dog_age: int) -> list:
     if (not isinstance(cat_age, (int, float))
             or not isinstance(dog_age, (int, float))):
-        raise TypeError
-    cat_age = int(cat_age)
-    dog_age = int(dog_age)
-    list_of_years = [15, 9, 4]
-    result = [0, 0]
+        raise TypeError("Type of age must be int or float")
+    dict_years = {"cat": [15, 9, 4], "dog": [15, 9, 5]}
```
petrykivd: It's not a good practice to use type names as variable names. In my opinion better is animal_lifespans or just lifespans (cause we know context is about animals)

app/main.py
```diff
@@ -1,30 +1,25 @@
...
-        else:
-            break
-        while (dog_age >= 0) and i == list_of_years[2]:
-            dog_age -= i
-            if dog_age >= 0:
-                result[1] += 1
+        while (age >= 0) and i == list_of_years[2]:
+            age -= i
+            if age >= 0:
+                result += 1
```
petrykivd, line 24: Maybe we can simplify solution and avoid loop? 😃

app/test_main.py
petrykivd: I think the test name needs improvement.

Naming a test `test_should_raise_error` is essentially the same as naming a function `do_something` — technically true, but completely uninformative.

A test function name should immediately tell us:
- **What** is being tested
- **What** the expected result is
- **When** / under what condition

A solid pattern to follow:


test_{what_is_tested}_{expected_result}_when_{condition}
For exampl,e :

test_get_human_age_raises_type_error_when_args_are_not_int


Think of the test name as documentation — someone should understand what the test does without reading its body.

**Also it applies to vairables/parameters names. Check it again please** 😉

petrykivd (changes requested):

app/test_main.py
pawunder1992: solved

app/main.py
pawunder1992: done

app/main.py
pawunder1992: I cant do this without AI)

VadimPusiak788 (approved):

PR #1979, opened 2026-04-06
'Solution'

artradkevych:


app/test_main.py
```diff
@@ -1,3 +1,26 @@
...
+        (23, 23, [1, 1]),
+        (24, 24, [2, 2]),
+        (28, 28, [3, 2]),
+        (100, 100, [21, 17]),
+    ]
+
+)
+def test_get_human_age(cat_age: int, dog_age: int, result: list) -> None:
+    assert get_human_age(cat_age, dog_age) == result
+
```
petrykivd, line 22: You can add test which should test TypeError raising

app/main.py
```diff
@@ -1,23 +1,13 @@
 def get_human_age(cat_age: int, dog_age: int) -> list:
```
petrykivd, line 1: What happens if I pass non-numeric values here?

petrykivd (changes requested):

petrykivd (approved):
Nice! 👍 
I want to emphasize that commit titles should reflect the code they describe. 

So simply calling it “Solution” isn't a good idea.

I recommend using titles like Fixed/Updated/Implemented/Added/Removed...

In this example: Added age parameter type checking and a new test...

PR #2015, opened 2026-04-19
solution: fix syntax error and pass all tests

RustamHadoiev24:


smykroman-dot (commented):
AI won’t accept it.
If function resive data out of normal range, negative numbers;
If the function receives an incorrect type of data it raises the correct exception.

Use ```@pytest.mark.parametrize```

antar1x (approved):

PR #2036, opened 2026-04-30
Solution

RWAdelP:


RWAdelP:
@mateacademy-ai-mentor Consistency is required. The check list used to be for good code practices and now it contains information that should be in the Readme.

PR #2049, opened 2026-05-07
Solution

swirtam:


swirtam:
@mateacademy-ai-mentor you say that for example "the test expects dog_human_age=3 for dog_age=31, but the correct calculation is 2 + (31-24)//5 = 3" and that's exactly what the test expects, for dog human age to be 3. Moreover, "dog age at 28 should convert to 2 human years" and that's what the test checks and that is the value returned by the function. "dog_human_age=2 for dog_age=27" and the function returns 2, and the function should not return one. The tests are trying to check mainly the borderline examples, like for cats 32 is a borderline example for 4 years, but for dogs it's 34, there is no point in checking 32 input for dog age since it's not a borderline example

swirtam:
@mateacademy-ai-mentor If the goal is to check if
* for a cat age = 28 the human age is 3
* plus for a dog age = 28 the human age is 2
* then these two inputs are checked  in lines 15-16. Tests (27, 28, 2, 2), (28, 29, 3, 3), are checking the case (28, 28, 3, 2) just not in a single test. I got the impression that the examples in the readme file are meant to showcase the function performance, not that they need to be replicated by testcases to the letter

PR #2106, opened 2026-06-16
Solution for Cat and Dog Years

maxmlv:


maxmlv:
@mateacademy-ai-mentor

> This violates checklist item: if function resive data out of normal range, such as negative numbers, zero, or realy large numbers;

So, if age parameter is Zero, the function should validate it as the correct value?

PR #2133, opened 2026-07-01
ss

golflang:


app/main.py
```diff
@@ -1,23 +1,34 @@
...
+        cat_age -= 24
+        cat_human_age = (cat_age // 4) + 2
+
+    if dog_age < 15:
+        dog_human_age = 0
+    elif 15 <= dog_age <= 23:
+        dog_human_age = 1
+    else:
+        dog_age -= 24
+        dog_human_age = (dog_age // 5) + 2
```
hladiivolodymyr: The same conditional logic is repeated for both cats and dogs. To follow the DRY (Don't Repeat Yourself) principle, consider extracting the shared logic into a helper function and passing the differing value (4 for cats and 5 for dogs) as a parameter.

hladiivolodymyr (changes requested):

hladiivolodymyr (approved):

PR #2144, opened 2026-07-08
sol

neokitoss:


vsmutok (changes requested):

app/main.py
```diff
@@ -1,23 +1,18 @@
...
+        cat_human_age = 2 + (cat_age - 24) // 4
+
+    if dog_age < 15:
+        dog_human_age = 0
+    elif dog_age < 24:
+        dog_human_age = 1
+    else:
+        dog_human_age = 2 + (dog_age - 24) // 5
+
+    return [cat_human_age, dog_human_age]
```
hladiivolodymyr: There’s some duplicated logic here. Since the calculation is the same for both cats and dogs, with only the divisor being different (4 for cats and 5 for dogs), you could extract it into a helper function and pass the divisor as a parameter. This would reduce duplication and make the code easier to maintain.

hladiivolodymyr (changes requested):

hladiivolodymyr (approved):

PR #2152, opened 2026-07-14
implement get_human_age func and tests

RuslanSamsonovych:


viktoria-rybenchuk (commented):
no need make changes in main.py. Please rollback by command
git checkout master -- app/main.py

viktoria-rybenchuk (changes requested):
read the comment

app/main.py
```diff
@@ -1,23 +1,20 @@
...
+            human_age.append(value // 15)
+        elif 16 <= value <= 24:
+            human_age.append(1 + (value - 15) // 9)
+        else:
+            if key == "cat_age":
+                human_age.append(2 + (value - 24) // 4)
+            if key == "dog_age":
+                human_age.append(2 + (value - 24) // 5)
+
+    return human_age
```
hladiivolodymyr: The solution works correctly, but it's more complex than necessary.

There's no real need to create a dictionary and iterate over it just to distinguish between cats and dogs. A cleaner approach would be to extract the shared logic into a helper function and pass the only varying parameter. This would eliminate the unnecessary dictionary, string comparisons, and nested conditionals, making the code shorter, easier to read, and easier to maintain.

hladiivolodymyr (changes requested):

hladiivolodymyr (approved):

PR #2171, opened 2026-07-30
Solution

rymmaprodan6-ops:


vsmutok (changes requested):
The tests have failed. 
Please make sure your code passes all tests before submitting the task for review.

app/main.py
```diff
@@ -1,23 +1,22 @@
...
+        if age < 24:
+            return 1
+        return 2 + (age - 24) // 4
+
+    def dog_to_human(age: int) -> int:
+        if age < 15:
+            return 0
+        if age < 24:
+            return 1
+        return 2 + (age - 24) // 5
```
hladiivolodymyr: The cat_to_human and dog_to_human functions contain duplicated logic. They differ only in the divisor used in the age calculation (4 for cats and 5 for dogs). Consider extracting the common logic into a single helper function and passing the divisor as a parameter

hladiivolodymyr (changes requested):

hladiivolodymyr (changes requested):
Please fix the previous comment, and after that, request a review

hladiivolodymyr (approved):

PR #2180, opened 2026-08-06
Solution

nalyvaiko-denys:


app/main.py
```diff
@@ -1,23 +1,16 @@
...
+        cat_human_age = 1
+    else:
+        cat_human_age = 2 + (cat_age - 24) // 4
+
+    if dog_age < 15:
+        dog_human_age = 0
+    elif dog_age < 24:
+        dog_human_age = 1
+    else:
+        dog_human_age = 2 + (dog_age - 24) // 5
```
hladiivolodymyr: This logic is duplicated. Consider extracting it into a helper function to improve readability and avoid repeating the same code

app/test_main.py
```diff
@@ -1,3 +1,37 @@
...
+def test_cat_gets_third_human_year_before_dog() -> None:
+    assert get_human_age(28, 28) == [3, 2]
+
+
+def test_dog_gets_third_human_year() -> None:
+    assert get_human_age(29, 29) == [3, 3]
+
+
+def test_large_ages() -> None:
+    assert get_human_age(100, 100) == [21, 17]
```
vsmutok: ## Clean Code

1. Add comments, prints, and functions to check your solution when you write your code. 
Don't forget to delete them when you are ready to commit and push your code.

2. Make sure you've tests checks all edged situations such as:
* if output should change with some integer value you should 
check that it is not changed with previous value;
* if function resive data out of normal range, 
such as negative numbers, zero, or realy large numbers;

3. Use ```@pytest.mark.parametrize```

